PR #17, opened 2022-10-31
Add runner, intermediate datasets, methods processing multiple inputs, and stats

yousefmoazzam:
The main changes are:
- add task runner
- add functionality to save intermediate datasets (global flag `--save_all` to save intermediate datasets for all method outputs, as well as `save_result` parameter for each method to save a single method's output)
- add functionality to have a method process multiple datasets using the `data_in_multi` and `data_out_multi` parameters
- pass stats to method functions that require them


yousefmoazzam:
Note the comment in 3f9cd69 which explains that the image saver currently breaks with the runner due to the `data_out` parameter being missing (during dataset initialisation).

We should probably figure out some way to standardise savers so that the runner can handle them correctly when it tries to go through a method's params and initialise datastes specified in `data_in` and `data_out`, since savers likely won't have a `data_out` parameter but something else instead?

yousefmoazzam:
> Note the comment in [3f9cd69](https://github.com/DiamondLightSource/HTTomo/commit/3f9cd6990d37dccbd5a2fe3c231def9bfceab4e3) which explains that the image saver currently breaks with the runner due to the `data_out` parameter being missing (during dataset initialisation).
> 
> We should probably figure out some way to standardise savers so that the runner can handle them correctly when it tries to go through a method's params and initialise datastes specified in `data_in` and `data_out`, since savers likely won't have a `data_out` parameter but something else instead?

No standardisation has been implemented as such, but a workaround is now in place to allow `httomo.misc.images.save_to_images` to work with the runner despite it not having a `data_out` or `data_out_multi` parameter. See [427d6b9](https://github.com/DiamondLightSource/HTTomo/pull/17/commits/427d6b9223a4272c41765e1214ba177fa5bf3068) for details on the workaround.

PR #30, opened 2022-11-17
centering methods preview and moving function to httomolib

dkazanc:


src/wrappers/tomopy/recon.py
```diff
@@ -58,13 +59,15 @@ def rotation(params: Dict, method_name:str, comm: Comm, data: ndarray) -> float:
...
     """
+   
     module = getattr(recon, 'rotation')
     method_func = getattr(module, method_name)
     rot_center = 0
     mid_rank = int(round(comm.size / 2) + 0.1)
     if comm.rank == mid_rank:
-        mid_slice = data.shape[1] // 2
-        rot_center = method_func(data[:, mid_slice, :], **params)
+        if params['ind'] == 'mid':
```
dkazanc, line 68: Actually we don't need a preview as such here as we've got **ind** parameter from TomoPy which specifies the index. I just enabled here also 'mid' to be the middle index of the dataset.

samples/pipeline_template_examples/02_basic_cpu_pipeline_tomo_standard.yaml
```diff
@@ -36,7 +36,7 @@
     find_center_vo:
       data_in: tomo
       data_out: cor
-      ind: null
```
dkazanc, line 39: @yousefmoazzam do you know why **ind** is generated as _null_ instead of _None_ as in [TomoPy function](https://tomopy.readthedocs.io/en/stable/api/tomopy.recon.rotation.html#tomopy.recon.rotation.find_center_vo)?

dkazanc:
@yousefmoazzam I've moved functions to httomolib in this PR, please check that the `preview` related changes makes sense. thanks!

PR #46, opened 2023-01-31
Restructuring project for use with pyproject.toml file only

team-gpu:
This does a similar restructuring as recently done on httomolib - using only pyproject.toml for all setuptools and tool configurations. It also puts it into a flat structure, avoiding `src`. 

Note that it also specifies 3 optional depedencies: 
- httomolib
- tomopy
- dev

The first two are basically the backends, in case users want to install `httomo` with just one of these backends (`pip install httomo[httomolib]`, etc.). 

The last one is for development, with testing tools + both tomopy and httomolib.


README.rst
```diff
@@ -11,21 +11,33 @@ Setup a Development Environment:
...
+* Install the module + backend(s) with :code:`pip install .[httomolib,tomopy]`
 
 Running the code:
 =================
 
 * Install the module as described in "Install as a Python module"
-* Execute the python module with :code:`python -m httomo <args>`
-* For help with the command line interface, execute :code:`python -m httomo --help`
+* Execute the python module with :code:`httomo <args>`
+* For help with the command line interface, execute :code:`httomo --help`
```
namannimmo10, line 28: This does not work for me.

pyproject.toml
```diff
@@ -0,0 +1,113 @@
...
+httomolib = [ "httomolib" ]
+dev = [
+  "pytest",
+  "pytest-cov",
+  "pytest-xdist",
+  "flake8",
+  "pyproject-flake8",
+  "pydocstyle",
+  "toml",
+  "httomolib[dev]",
```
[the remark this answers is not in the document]
lotzej: I don't understand... I don't think that's a typo, what do you mean?

README.rst
lotzej: After installing, it should work. It adds a script to the bin folder for httomo. It might be that the bin folder is not in your path? 

We can look at it together tomorrow.

Issue #40, opened 2023-01-18
Implement multiple parameter values for methods in task runner


yousefmoazzam:
This should reproduce the "parameter tuning" feature in Savu, where a parameter for a method could be given multiple values and the method would be applied multiple times to its input data with those different parameter values. The output data of the pipeline would reflect these multiple parameter values by having an extra dimension, in which the different elements would correspond to the different parameter values.

yousefmoazzam:
As explained in #51, the output file of a parameter sweep no longer has a higher dimensionality than the individual results. Instead there is a single hdf5 file with multiple datasets within that file, and each dataset in the file represents the result of a particular value in the parameter sweep.

yousefmoazzam:
Implemented in #62

PR #62, opened 2023-02-14
Add parameter sweep functionality [WIP]

yousefmoazzam:
The main feature to be added in this PR is the ability to perform "parameter sweeps" as described in #40. There is also some refactoring of the main loop in the task runner.

There are two ways in which a parameter sweep can be defined:
- defining a range via start, stop, step values
- explicitly listing all values to sweep across 

and there are two example pipeline YAML configs showing how to do each in https://github.com/DiamondLightSource/httomo/tree/param-sweeps/samples/pipeline_template_examples/parameter_sweeps.

The following situations are currently handled, but there may be more that have not yet been encountered:
- the method with the parameter sweep is the last one in the pipeline
- the method with the parameter sweep is after a reslice operation
- the method with the parameter sweep is before a reslice operation


yousefmoazzam:
@dkazanc I had an issue with saving middle slices that seemed to require some fiddling with MPI, so note that d2fe7ef has some MPI-related code.

When running with multiple processes, the data is split between the processes, which means that each process only has access to a subset of the data and not the entire "global data". The issue I came across was that this meant that only one of the MPI processes had access to the middle slice, and the rest did not.

However, it didn't appear to be trivial to figure out which MPI process was the one that had access to the middle slice (due to the data potentially being split unequally among the processes if the length of the axis to split along is not divisible by the number of processes).

The code that I added in this commit is effectively:
- using [`allgather`](https://mpi4py.readthedocs.io/en/stable/reference/mpi4py.MPI.Comm.html#mpi4py.MPI.Comm.allgather) to have each MPI process notify every process what the length of its subset of data is (along the splitting-axis)
- doing some arithmetic in each process (using the lengths of the split data from other processes) to decide if it has access to the "global" middle slice or not
- only the MPI process with access to the "global" middle slice is fetching the middle slices, and saving tiffs of the middle slices

I'm not at all experienced with MPI so this could easily not be a good way to do this, so please feel free to change that if a better way is available!

dkazanc (approved):

PR #70, opened 2023-02-22
360 degrees pipeline

dkazanc:
Here are corrections to make 360 degrees pipeline run. 
The most significant changes are the following ones:
- task runner to enable dealing with [multiparameters output](https://github.com/DiamondLightSource/httomo/blob/089c5f3ac969cd36e6d366483ff17a6e5b39581e/samples/pipeline_template_examples/DLS/02_i12_360scan_pipeline.yaml#L23) in yamls. By default yaml will convert that into a tuple so I'm dealing with it in two stages in the runner: first, [I preserve the list with names while also storing the tuple](https://github.com/DiamondLightSource/httomo/blob/089c5f3ac969cd36e6d366483ff17a6e5b39581e/httomo/task_runner.py#L221); secondly [I pass the list down to the runner and then re-assign values from the tuple in datasets](https://github.com/DiamondLightSource/httomo/blob/089c5f3ac969cd36e6d366483ff17a6e5b39581e/httomo/task_runner.py#L509). I guess this is just a one way of doing it.
- Had to tweak few wrappers and especially important to truncate the angles to [fit](https://github.com/DiamondLightSource/httomo/blob/089c5f3ac969cd36e6d366483ff17a6e5b39581e/wrappers/httomolib/recon.py#L50) the size of the truncated data. 

re: test. The new [pipeline](https://github.com/DiamondLightSource/httomo/blob/089c5f3ac969cd36e6d366483ff17a6e5b39581e/samples/pipeline_template_examples/DLS/02_i12_360scan_pipeline.yaml) can be applied to the stadard data. It should work but produce nonsense. 

Reconstructed circle means the success, but a bit boring sample I have to say! 
image


yousefmoazzam (commented):
This all looks fine to me, both @dkazanc and I have tested this with `tomo_standard.nxs`, and @dkazanc has also tested this with the real data shown as an image in the PR.

The only other thing to add maybe is the reduction of the `angles` array from 360 degrees down to 180 degrees when the data has been modified by the stitching. However, this is not needed to have the 360 pipeline work, only for getting correct metadata about the rotation angles, which could be left for later if necessary.

dkazanc:
fair point on angles, so far there is a little [fix](https://github.com/DiamondLightSource/httomo/blob/16a3632420ead22cf872c15f2f3e53d38f6c4d3a/wrappers/httomolib/recon.py#L48) in the reconstruction wrapper, but I might deal with it as well when I will be doing the wrapper class.

PR #72, opened 2023-02-23
Setting up GitHub Actions for CPU tests

namannimmo10:


conda/environment.yml
```diff
@@ -8,6 +8,7 @@ dependencies:
   - astra-toolbox::astra-toolbox
   - conda-forge::click
   - conda-forge::cupy
+  - conda-forge::cudatoolkit=11.0
```
namannimmo10, line 11:
```suggestion
```

namannimmo10:
We can run the CPU pipeline integration tests on GitHub Actions as we sort out the CuPy-related import errors.
The CI is green (currently runs only _test_cli.py_) because it is not yet running the CPU pipeline with the `httomolib` wrappers... and if it does, you'd see an error, 
> `ModuleNotFoundError: No module named 'httomolib'`

And that's because `httomolib` is not in the environment. The next step is to either conda-forge `httomolib` or make it available via `savu-dep` - whichever works.

.github/workflows/cpu_tests.yml
```diff
@@ -0,0 +1,39 @@
...
+      - name: Setup Python 3.9
+        uses: actions/setup-python@v2
+        with:
+          python-version: 3.9
+
+      # install dependencies with conda
+      - name: Install dependencies with Conda
+        run: |
+          conda env create httomo --file conda/environment.yml
+          conda run -n httomo pip install -e .
```
namannimmo10, line 33:
```suggestion
          conda run -n httomo pip install -e .
          conda list
```

PR #74, opened 2023-03-02
A class for wrappers [WIP]

dkazanc:


httomo/wrappers_class.py
```diff
@@ -0,0 +1,170 @@
...
+        # for 360 degrees data the angular dimension will be truncated while angles are not.
+        # Truncating angles if the angular dimension has got a different size
+        angular_dim_size = np.size(data, 0)
+        if np.size(data, 0) != len(angles_radians):
+            angles_radians = angles_radians[0:angular_dim_size]
+                
+        data = getattr(self.module, method_name)(data, angles_radians, **params)
+        return data   
+    
+class tomopy_wrapper(wrappers):
```
namannimmo10, line 93: class names are CamelCase: https://google.github.io/styleguide/pyguide.html#3164-guidelines-derived-from-guidos-recommendations
```suggestion
class TomoPyWrapper(wrappers):
```

httomo/wrappers_class.py
```diff
@@ -0,0 +1,170 @@
...
+            self.gpu_id = int(comm.rank / comm.size * self.num_GPUs)
+        if module_name == "misc":
+            from httomolib import misc
+            self.module = getattr(misc, function_name) # imports the module
+        elif module_name == "prep":
+            from httomolib import prep
+            self.module = getattr(prep, function_name) # imports the module
+        elif module_name == "recon":
+            from httomolib import recon
+            self.module = getattr(recon, function_name) # imports the module
```
namannimmo10, line 145: I'd handle imports at the top of the file or use `import_module`.
```suggestion
        if module_name in ["misc", "prep", "recon"]:
            from importlib import import_module
            self.module = getattr(import_module("httomolib." + module_name), function_name)
```

httomo/wrappers_class.py
```diff
@@ -0,0 +1,191 @@
...
+        # if not changed bellow the generic wrapper will be executed
+        self.wrapper_method = super(TomoPyWrapper, self)._execute_generic
+
+        if module_name in ["misc", "prep", "recon"]:
+            from importlib import import_module
+            self.module = getattr(import_module("tomopy." + module_name), function_name)
+            if function_name == "normalize":
+                func = getattr(self.module, method_name)
+                sig_params = signature(func).parameters
+                if 'dark' in sig_params and 'flat' in sig_params:
```
namannimmo10, line 139:
```suggestion
                if 'darks' in sig_params and 'flats' in sig_params:
```

httomo/wrappers_class.py
```diff
@@ -0,0 +1,191 @@
...
+                    self.wrapper_method = super(TomoPyWrapper, self)._execute_normalize                
+            if function_name == "algorithm":
+                self.wrapper_method = super(TomoPyWrapper, self)._execute_reconstruction
+            if function_name == "rotation":
+                self.wrapper_method = super(TomoPyWrapper, self)._execute_rotation
```
namannimmo10, line 144: I'd add a dictionary to `BaseWrapper` that maps the `module` to the method which I'd like to execute,
```python
_module_to_method = {
    "algorithm": _execute_reconstruction,
    "rotation" : _execute_rotation,
    "normalize": _execute_normalize
}
```
The common logic, i.e. checking `darks` and `flats` in the parameters can also be moved to  `BaseWrapper`.
So, in a way, all of this can be done in one `Wrapper` class. But if you want to keep it clean and separate, two classes make sense. Because down the road, we can add wrappers for other new packages like `scipy`, etc.

httomo/wrappers_class.py
dkazanc: TomoPy [uses](https://tomopy.readthedocs.io/en/stable/api/tomopy.prep.normalize.html#tomopy.prep.normalize.normalize) `dark` and `flat` naming somehow.

namannimmo10:
To make the linter a happy bunny, please do
```console
$ pip install black
$ black .
```

httomo/wrappers_class.py
```diff
@@ -0,0 +1,237 @@
...
+                func = getattr(self.module, method_name)
+                sig_params = signature(func).parameters
+                if "dark" in sig_params and "flat" in sig_params:
+                    self.wrapper_method = super(
+                        HttomolibWrapper, self
+                    )._execute_normalize
+            if function_name == "algorithm":
+                self.wrapper_method = super(
+                    HttomolibWrapper, self
+                )._execute_reconstruction
```
namannimmo10, line 183:
```suggestion
                self.wrapper_method = super()._execute_reconstruction
```

dkazanc: thanks, this simplifies the look.

httomo/wrappers_class.py
```diff
@@ -0,0 +1,237 @@
...
+            pass
+
+        if gpu_enabled:
+            data = getattr(self.module, method_name)(
+                xp.asnumpy(data), out_dir, comm_rank=comm.rank, **params
+            )
+        else:
+            data = getattr(self.module, method_name)(
+                data, out_dir, comm_rank=comm.rank, **params
+            )
```
namannimmo10, line 216:
```suggestion
        for key in ["ncore", ""]:  #: silently remove `ncore` param.
            params.pop(key, None)
        data = getattr(self.module, method_name)(
            data.get() if gpu_enabled else data, out_dir, comm_rank=comm.rank, **params
        )
```

httomo/wrappers_class.py
```diff
@@ -0,0 +1,237 @@
...
+try:
+    import cupy as xp
+
+    try:
+        xp.cuda.Device(0).compute_capability
+        gpu_enabled = True  # CuPy is installed and GPU is available
+    except xp.cuda.runtime.CUDARuntimeError:
+        import numpy as xp
+
+        print("CuPy is installed but GPU device inaccessible")
```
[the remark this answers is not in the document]
namannimmo10: I am guessing you are using this attribute to check if `Device(0)` is accessible -- makes sense.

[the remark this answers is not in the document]
team-gpu: The point is, you wanna try making one call to the CUDA runtime library to make sure it is possible, and there are administrative settings etc that can also block you from using a GPU. All of these that I'm aware of would limit the device count visible to the process, so checking device count is likely enough. 

For example, an admin might block the process from using any GPU, in which case the `getDeviceCount()` call would just return 0.

Of course there is the case that the GPU is somehow busy or faulty - if that's the case, neither method will work and it would simply crash further down the line. But that's rare enough and I wouldn't try to guard against that. (We don't check whether CPU RAM is working or cores are working before using them either)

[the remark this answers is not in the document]
dkazanc: yes, I do aim to get the CUDARuntimeError if there is no device 0 available. The zeroth device should be present as a starting point requirement to use a GPU. The number of devices is a minor importance at the stage I think...

httomo/wrappers_class.py
dkazanc: thanks, not quite sure that this does as I need to convert possibly a cupy array to a numpy one if the gpu is enabled.
```
data = getattr(self.module, method_name)(
            data.get() if gpu_enabled else data, out_dir, comm_rank=comm.rank, **params
        )
```

namannimmo10: `.get()` is a way to get a copy of the ndarray on host memory: https://docs.cupy.dev/en/stable/reference/generated/cupy.ndarray.html#cupy.ndarray.get

namannimmo10: Either will work!

dkazanc:
OK, I think I've tested it enough and it does seem to be working. I've also fixed some inconsistencies we had with `ncore` and `gpu_id` parameters being passed. **Please test and report!** When tested, please use the changes in httomolib in the [wrapper](https://github.com/DiamondLightSource/httomolib/pull/80) branch.

httomo/wrappers_class.py
```diff
@@ -0,0 +1,308 @@
...
+class BaseWrapper:
+    """A parent class for all wrappers in httomo that use external modules."""
+
+    def __init__(
+        self, module_name: str, function_name: str, method_name: str, comm: Comm
+    ):
+        self.comm = comm
+        if gpu_enabled:
+            self.num_GPUs = xp.cuda.runtime.getDeviceCount()
+            self.gpu_id = int(comm.rank / comm.size * self.num_GPUs)
```
team-gpu, line 35: I don't think this is right - assume there are 8 processes spread over 2 nodes with 4 GPUs each. You'd get:

| rank | node | gpu_id |
|------|------|--------|
| 0 | 0 | 0 |
| 1 | 0 | 0 |
| 2 | 0 | 1 |
| 3 | 0 | 1 |
| 4 | 1 | 2 |
| 5 | 1 | 2 |
| 6 | 1 | 3 |
| 7 | 1 | 3 |

httomo/wrappers_class.py
```diff
@@ -0,0 +1,308 @@
...
+                with xp.cuda.Device(self.gpu_id):
+                    # move the data to a particular device
+                    data = xp.asarray(data)
+                    flats = xp.asarray(flats)
+                    darks = xp.asarray(darks)
+            else:
+                # the method doesn't accept CuPy arrays
+                data = xp.asnumpy(data)
+                flats = xp.asnumpy(flats)
+                darks = xp.asnumpy(darks)
```
team-gpu, line 108: how about adding a utility method `def transfer_data(self, *args) -> tuple:` which does all the checking and conversion, to be called like:

```python
data, flats, darks = self.transfer_data(data, flats, args)
```

team-gpu (changes requested):

httomo/wrappers_class.py
dkazanc: yep, good idea. I can try that, thanks!

httomo/wrappers_class.py
dkazanc: Sorry I'm not sure how this is calculated. Say you've got 4 GPUs and 4 Processes then size.comm = 4 and you'll get 1,2,3,4 gpu indices for each process. The same distribution will be for a different node. The indexing of GPUs is local to the node, isn't it?

yousefmoazzam: I _think_ this is how the table has been constructed...

The values in the `gpu_id` column represent the different values of `self.gpu_id` in the code, which are calculated based on the different values of `comm.rank` for each MPI process via:
```python
self.gpu_id = int(comm.rank / comm.size * self.num_GPUs)
```

The values in the `rank` column are being substituted as `comm.rank` in the code, and then we set `comm.size=8` and `self.num_GPUs=4` for this particular example of
- 2 nodes
- 4 GPUs per node

which comes to 8 processes altogether.

If you plug those values into the line
```python
self.gpu_id = int(comm.rank / comm.size * self.num_GPUs)
```
for each of the 8 MPI processes (ie, changing `comm.rank`), you get the following:
```python
>>> int(0 / 8 * 4)
0
>>> int(1 / 8 * 4)
0
>>> int(2 / 8 * 4)
1
>>> int(3 / 8 * 4)
1
>>> int(4 / 8 * 4)
2
>>> int(5 / 8 * 4)
2
>>> int(6 / 8 * 4)
3
>>> int(7 / 8 * 4)
3
```

team-gpu: yes, indexing of GPUs is local, but the comm.size and comm.rank are global. So we need to find which of the ranks are on the same node (just different cores), and establish a local rank between them. So in the table above, it should restart at 0 for the global rank 4.

team-gpu: One example to do that is to establish a dict of rank -> hostname + gather it from all processes, and then count the ranks within the same hostname. Like here: https://github.com/ptycho/ptypy/blob/master/ptypy/utils/parallel.py#L746  ... that's rather complex, and I think this would work better:

```python
        local_comm = comm.Split_type(MPI.COMM_TYPE_SHARED)
        if gpu_enabled:
            self.num_GPUs = xp.cuda.runtime.getDeviceCount()
            self.gpu_id = local_comm.rank % self.num_GPUs
```

This uses a local communicator, splitting the global one into those that can communicate with shared memory. That is the same as the ranks on the same node. And within these groups, the rank count starts at 0 again. 

I've added the modulo, so that if for some reason you're putting 8 processes on a node with 4 GPUs, it would wrap around and put the rank 4 back on gpu 0. Another option is to throw an exception for that case.

yousefmoazzam: ...That's another bit of MPI magic that I've never seen before, thanks for pointing out the existence of `comm.Split_type()`!

team-gpu: Just had a short session with @dkazanc - I think it's best to create an MPI utility module in the codebase, which exports `rank`, `size`, `local_rank`, and `local_size`, and have a function in there which determines the local values. So it should be this for the different MPI processes, assuming 2 procs/node:


| size  | 4 | 4 | 4 | 4 |
|----|---|---|---|---|
|  rank  | 0 | 1 | 2 | 3 |
| local_size | 2 | 2 | 2 | 2 |
| local_rank | 0 | 1 | 0 | 1 |

It may be implemented using the `Split_type` call above.

And then everywhere else in the code, you use this utility to get the local ranks and sizes.

httomo/wrappers_class.py
```diff
@@ -0,0 +1,312 @@
...
+        for datasets in args:
+            if gpu_enabled:
+                xp.cuda.Device(self.gpu_id).use()
+                if self.cupyrun:
+                    # the method accepts CuPy arrays for the GPU processing
+                    # move the data to the current device
+                    ret = ret + (xp.asarray(datasets),)
+                else:
+                    # the method doesn't accept CuPy arrays
+                    ret = ret + (xp.asnumpy(datasets),)
```
team-gpu, line 55: and what if `gpu_enabled` is `False`?

team-gpu: but you're returning an empty tuple in that case...

dkazanc: oops!

httomo/wrappers_class.py
```diff
@@ -0,0 +1,307 @@
...
+            xp.ndarray: A numpy or cupy array containing processed data.
+        """
+        # set the correct GPU ID if it is required
+        if "gpu_id" in params:
+            params["gpu_id"] = self.gpu_id
+
+        # check where data needs to be transfered host <-> device
+        data = self._transfer_data(data)
+
+        data = getattr(self.module, method_name)(data[0], **params)
```
team-gpu, line 74: wondering why `data[0]` here? Would that work, considering the `transfer_data` above?

dkazanc: it doesn't like the `tuple` being passed...

yousefmoazzam: Doesn't `_transfer_data()` return a `tuple` of arrays (one array for each "output dataset" that a method returns I think): https://github.com/DiamondLightSource/httomo/blob/ffb0610047261a23b84090afeee0218cbdbd2d18/httomo/wrappers_class.py#L40-L51

So if `data` becomes a `tuple` of arrays at line 72:
```python
data = self._transfer_data(data)
```
then `data[0]` would be ok?

team-gpu: Ah, it gets a tuple with one element now of course. Perhaps we can adjust the `transfer_data` function to keep the tuple as a local variable, say `ret`, and at the end: `return ret[0] if len(ret) == 1 else ret`

dkazanc: implemented as suggested, thanks!

PR #78, opened 2023-03-03
Run CPU tests on the CI

namannimmo10:


team-gpu:
Regarding the tests that need httomolib - I guess if you want to test CPU only, you'll have to exclude those (now that we decided httomolib is GPU only). An alternative to test the wrapper would be using mocks for the httomolib functions, which export the same interface and docstrings, but don't do anything useful. That way you can test the wrappers on their own. Let me know if you need help with mocks (`pytest-mock` is the package to use).

namannimmo10:
Can try mocking, but these tests should be able to run without requiring a GPU because the `httomolib` wrapper in the testing pipeline uses only `save_to_images` which [does not require cupy](https://github.com/DiamondLightSource/httomolib/blob/377dc6b32469baf1c287853fcdd9e472ab818ce7/httomolib/misc/images.py#L36).
But yes, obviously we need to separate CPU and GPU pipeline integration tests because we cannot run all tests on GitHub Actions.

team-gpu:
> Can try mocking, but these tests should be able to run without requiring a GPU because the `httomolib` wrapper in the testing pipeline uses only `save_to_images` which [does not require cupy](https://github.com/DiamondLightSource/httomolib/blob/377dc6b32469baf1c287853fcdd9e472ab818ce7/httomolib/misc/images.py#L36). But yes, obviously we need to separate CPU and GPU pipeline integration tests because we cannot run all tests on GitHub Actions.

I understand it should work, but then you need to install httomolib in this CI pipeline.

.github/workflows/cpu_tests.yml
```diff
@@ -0,0 +1,43 @@
...
+          mamba activate httomo
+          mamba run -n httomo pip install -e .
+          mamba run -n httomo pip install httomolib
+          mamba list
+
+      # run tests
+      - name: Run CPU tests
+        run: |
+          mamba run -n httomo pip install pytest
+          mamba run -n httomo pytest
```
namannimmo10, line 43:
```suggestion
          conda install -c conda-forge mamba
          mamba env create --name httomo --file conda/environment.yml
          mamba activate httomo
          conda run -n httomo pip install -e .
          conda run -n httomo pip install httomolib
          conda list

      # run tests
      - name: Run CPU tests
        run: |
          conda run -n httomo pip install pytest
          conda run -n httomo pytest
```

PR #82, opened 2023-03-09
Reslice Optimisation using MPI (several minutes down to less than 1 second)

team-gpu:
## Overview

This PR optimises the `reslice` function using MPI directives rather than going through h5 files on disk. This has a dramatic effect on performance and works within reasonable memory.

## How it works

1. Every process builds a list of the part of its data that it needs to distribute to the other processes. That's a list of slices of the data that it has.
2. Using MPI `alltoall` call, every process sends this list to every other process in the communicator. The result is that each process gets a list of array slices it needs for the new slicing dimension.
3. Each process is concatenating this list, i.e. the bits it got from the other processes, into a single data array.

Also, if there is only one process, it does not do anything (no re-slicing is needed).

## Tests and Memory Consumption

This approach passes all the tests with all slicing dimensions and it doesn't need a file for intermediate data. It requires one copy of the data, so roughly twice the CPU memory that the data already has. It is reasonable to assume that the CPU has this amount of memory available in all cases, otherwise many other methods would fail as well.

## Performance

We've profiled this compared to the original code, using projection -> sinogram conversion and the other way around. Both show approximately the same performance characteristics, so we only show the former below. Also, for the old code, we used `/tmp` as well as a location in `/home` for the intermediate file, to see the effect of using a network-mounted drive for the intermediate files (which would be required in production). The tests have been run on a cluster node, with these results:

|    MPI Proc. | Size per Proc. |    Total |    /tmp (ms) | /home (ms) |     New (ms)    | New vs /tmp | New vs /home |
|---:|---:|---:|---:|---:|---:|---:|---:|
|     1 |     263MB |     263MB |            1,076  |          535,921  |                         0  | 67,250x | 33,495,044x |
|     2 |     263MB |     527MB |              3,559  |            820,163  |                       657  | 5x | 1,248x |
|     3 |     263MB |     791MB |            5,484  |          985,310  |                       656  | 8x | 1,502x |
|     4 |     263MB |     1054MB |              7,486  |        1,203,856  |                       646  | 12x | 1,864x |
|     5 |     263MB |     1319MB |            9,163  | 2,303,832 |                       645  | 14x | 3,572x |
|     6 |     263MB |     1581MB |            11,085  | 2,309,642 |                       710  | 16x | 3,253x |

Note:
- This new version is almost constant time, no matter how many processes are used
- The old times will vary with disk and network speed and should be seen as indicative only


team-gpu:
Note that for compatibility, the new reslice function is still taking the output directory as a parameter. I guess this should be removed from the function as well as the wrapper code that calls it - is that ok?

yousefmoazzam:
Yep, that makes sense to remove the `run_out_dir` parameter of the `reslice()` function along with this change to the reslicing functionality.

The only thing is to make sure to change everywhere that uses the `reslice()` function; aside from `task_runner.py`, I think there are several Jupyter notebooks in the repo that use it, like this one https://github.com/DiamondLightSource/httomo/blob/main/samples/jupyter_notebooks/DLS_specific/01_sino360_to180convert_inpainting.ipynb.

dkazanc:
> Yep, that makes sense to remove the `run_out_dir` parameter of the `reslice()` function along with this change to the reslicing functionality.
> 
> The only thing is to make sure to change everywhere that uses the `reslice()` function; aside from `task_runner.py`, I think there are several Jupyter notebooks in the repo that use it, like this one https://github.com/DiamondLightSource/httomo/blob/main/samples/jupyter_notebooks/DLS_specific/01_sino360_to180convert_inpainting.ipynb.

Don't worry about the notebooks, they need to be re-written...

team-gpu:
> > Yep, that makes sense to remove the `run_out_dir` parameter of the `reslice()` function along with this change to the reslicing functionality.
> > The only thing is to make sure to change everywhere that uses the `reslice()` function; aside from `task_runner.py`, I think there are several Jupyter notebooks in the repo that use it, like this one https://github.com/DiamondLightSource/httomo/blob/main/samples/jupyter_notebooks/DLS_specific/01_sino360_to180convert_inpainting.ipynb.
> 
> Don't worry about the notebooks, they need to be re-written...

I have changed them all now as well

namannimmo10:
Thanks, @team-gpu . It would be nice to add some pipeline integration tests for these:

> - `httomo <infile> <yaml> <outdir> task_runner` does the reslicing in-memory, using the optimised MPI directives in this PR
> - `httomo --file-based-reslice <infile> <yaml> <outdir> task_runner` uses files for reslicing (the old way), placed into `<outdir>`
> - `httomo --file-based-reslice --reslice-dir /tmp <infile> <yaml> <outdir> task_runner` uses files for reslicing, placed into `/tmp`

PR #89, opened 2023-03-15
raises errors for incorrect preview parameters

dkazanc:
the flashing message in red and the raised error. 
image


dkazanc:
will be merged with #95

Issue #67, opened 2023-02-20
Incorrect gathering of data in MPI runs with non-recon methods with sinogram pattern at end of pipeline


yousefmoazzam:
The cause of this is the same as in #66, just that it's present in saved files other than parameter sweeps.

The pipeline used was the following:
```yaml
- httomo.data.hdf.loaders:
    standard_tomo:
      name: tomo
      data_path: /entry1/tomo_entry/data/data
      image_key_path: /entry1/tomo_entry/instrument/detector/image_key
      dimension: 1
      pad: 0
- tomopy.prep.normalize:
    normalize:
      data_in: tomo
      data_out: tomo
      cutoff: null
- tomopy.prep.normalize:
    minus_log:
      data_in: tomo
      data_out: tomo
- tomopy.prep.stripe:
    remove_all_stripe:
      data_in: tomo
      data_out: tomo
      snr: 3
      la_size: 61
      sm_size: 21
      dim: 1
```

and was run with two MPI processes on `tomo_standard.nxs`.

Here is the result of the final method, and note the incorrect dimensions of the data (double the original number of projections, and half the original vertical height):

![im-1](https://user-images.githubusercontent.com/13002919/220063829-43f3d900-d02e-47de-a7d5-25a7bebcf0a4.png)

![im-2](https://user-images.githubusercontent.com/13002919/220063857-ddde5992-3dc2-449d-a03e-954b1e3938bb.png)

Using zero-based indexing for the axes of an array and one-based indexing for the value of `slice_dim`, what happens is:
- the input data to the stripe removal method is split between the two MPI processes along axis 1 (sinogram dimension)
- data is processed by both MPI processes
- when saving the data, the data _should_ be gathered up along axis 1
- however, `slice_dim` in the `intermediate_dataset()` function is hardcoded to be `1`
https://github.com/DiamondLightSource/httomo/blob/905c5f4dfec4f2a70d65e46fab7c136c02bc1ebb/httomo/data/hdf/_utils/save.py#L48
- in the `save_data_parallel()` function that the `intermediate_dataset()` call eventually gets to, when `slice_dim=1`, the data is gathered along axis 0 (projection dimension) (remember, the `slice_dim` value uses one-based indexing, whereas numpy arrays use zero-based indexing, hence, the discrepancy between the two)
[...]

PR #94, opened 2023-03-22
Fix incorrect gathering of data when saving intermediate dataset

yousefmoazzam:
See issue #67 for more information on the bug being addressed.


yousefmoazzam:
I won't run `black` to fix the CI errors, since that'll modify lots of code that has nothing to do with the small change/fix in this PR, hope that's ok.

dkazanc (approved):
looks good, thanks for sorting it!

Issue #64, opened 2023-02-15
Global stats is not passed through to `save_to_images`


yousefmoazzam:
Currently, the `glob_stats` associated with the input to `save_to_images` is not passed through. This is due to the wrapper function whose signature is inspected by the httomo runner https://github.com/DiamondLightSource/httomo/blob/3d7b153d324dad8ba04de0287a6592868989793e/httomo/task_runner.py#L220-L221 not containing a `glob_stats` parameter https://github.com/DiamondLightSource/httomo/blob/3d7b153d324dad8ba04de0287a6592868989793e/wrappers/httomolib/misc.py#L11

This doesn't result in an error during the run due to `save_to_images()` performing its own stats calculations if `glob_stats=None` https://github.com/DiamondLightSource/httomolib/blob/01623d2211e654ec7073630e4ce698d6ff83965a/httomolib/misc/images.py#L97-L103. Nevertheless, this should still be considered unintended behavior.

PR #98, opened 2023-03-24
global stats corrections and fix for #64

dkazanc:
Also dealt with the problem of potentially passing CuPy arrays into calculation of global statistics


dkazanc:
Changed how data can be [found](https://github.com/DiamondLightSource/httomo/blob/edd13d61f707c0e1e7b377325cc9e6964793c38f/tests/test_pipeline.py#L75) in tests as previously it wasn't dynamic. So tests now working...

PR #106, opened 2023-04-06
GPU Memory workaround work httomo [wip]

dkazanc:
Working in conjunction with this [PR](https://github.com/DiamondLightSource/httomolib/pull/89) from httomolib


dkazanc:
Following the latest discussion we've got plenty of things to work on. 
- [ ] The change of the dimension issue for memory estimators in httomolib. Currently they [return](https://github.com/DiamondLightSource/httomolib/blob/075d941769ac06212b51d3a3b9d3acdaabe3c67f/httomolib/prep/normalize.py#L51) the number of max slices to fit into memory and the data type. However, this is based on the assumption that the shape of the data remains the same. It is not always true, for instance, the reconstructor. So we should also return the shape, so it can be passed to the next method for the memory estimation.
- [ ] Loader is a part of the [platform section loop](https://github.com/DiamondLightSource/httomo/blob/550f6da56e84fbdf9dab7b722e8fe525b76c8ee6/httomo/task_runner.py#L233) and should be moved away as it always the first method in the pipeline.
- [ ] Check patterns "all", "sino", "proj" consistency with the `reslice_ahead` and see if it somehow can be  generilised? Should the `all` pattern inherit the previous pattern explicitly in the runner? 
- [ ] TomoPy doesn't always provide the operating data type for a particular method, worth checking if we need to add this into library?  
- [ ] Refactoring of the code and make it more readable and structured BEFORE implementing the next point.
- [ ] A loop over data chunks inside the [platform section loop](https://github.com/DiamondLightSource/httomo/blob/550f6da56e84fbdf9dab7b722e8fe525b76c8ee6/httomo/task_runner.py#L233). It has been suggested that we should create a runner class where this can be a method? 
@yousefmoazzam and @namannimmo10 I suggest we decide who takes which part and we can link the separate developments to the list?

namannimmo10:
I can start with task 2 in the list. The diff in this PR is massive, we should try to split it into different PRs if possible.

dkazanc:
currently this branch fails with the issue fixed with this [PR](https://github.com/DiamondLightSource/httomo/commit/c1f393598d538b3e376eeeb2f51f95aa31afea14). Couple of options: 1. A fast fix as in the aforementioned PR. 2. A fix with the refactoring. If someone on refactoring already, worth considering that?

httomo/data/hdf/loaders.py
```diff
@@ -9,19 +11,43 @@
...
+
+@dataclass
+class LoaderData:
+    data: ndarray
+    flats: ndarray
+    darks: ndarray
+    angles: ndarray
+    angles_total: int
+    detector_x: int
+    detector_y: int
```
namannimmo10, line 33: same copy as above. 
```suggestion
```

httomo/task_runner.py
```diff
@@ -40,6 +43,98 @@
...
+    has_warn_printed : bool
+        Whether the reslicing warning has been printed
+    reslice_dir : Optional[Path]
+        The directory to use with file-based reslicing. If None,
+        reslicing will be done in-memory.
+    """
+
+    count: int
+    has_warn_printed: bool
+    reslice_dir: Optional[Path] = None
```
namannimmo10, line 107: rebase with `main` after #133  is in.

namannimmo10, line 107:
```suggestion
```

httomo/data/hdf/loaders.py
namannimmo10: actually this dataclass is added in #133, please rebase.

httomo/wrappers_class.py
```diff
@@ -1,12 +1,12 @@
+from typing import Any, Callable, Dict, Tuple, Union
+import numpy as np
 import inspect
-from inspect import signature
-from typing import Dict, Union
+from inspect import Parameter, signature
+from httomo.utils import Colour, log_once
```
namannimmo10, line 5:
```suggestion
from httomo.utils import Colour, log_exception, log_once
```

namannimmo10:
tests are failing because we need to upload the latest httomolib on the conda channel to fetch meta data from the decorators.

dkazanc:
yes, the decorators have been removed from httomolib and they are only in httomolibgpu now. Creating a library file should solve the issue with the CPU tests. I'll try to add it now

dkazanc:
removed newer builds for httomolib but now it fails on the cupy import from (older gpu-cpu) httomolib. I could try making this branch to work with httomolib and httomolibgpu recent changes or we prefer to merge it to main first as is?

httomo/wrappers_class.py
```diff
@@ -274,36 +281,19 @@ def __init__(
...
-
-        # httomolib can include GPU/CuPy methods as as well as the CPU ones. Here
-        # we check if the method can accept CuPy arrays by looking into docstrings.
 
-        # get the docstring of a method in order to check the I/O requirements
-        get_method_docs = inspect.getdoc(getattr(self.module, method_name))
-        # if the CuPy array mentioned in the docstring then we will enable
-        # the GPU run when it is possible
+        # httomolib methods run on the CPU only
         self.cupyrun = False
```
namannimmo10, line 296: `self.cupyrun` is `False` already from the `BaseWrapper`, we don't have to assign it to False in the derived classes.

PR #110, opened 2023-04-13
Memory chunking in httomo

team-gpu:
This adds memory chunking for sections of GPU code with the same pattern to httomo. In particular:

- It parses the pipeline and separates it into platform sections, i.e. sequences of the pipeline that run on the same platform and don't have a reslice in the middle
- Within these sections, for those that are running on GPU only, it determines the maximum number of slices that it can fit into the available GPU memory
- Then it iterates through the methods in these sections, processing the data in chunks as needed.

Further, it refactors the task runner to use data classes and clean up the parameter passing and method descriptions, as this is necessary for a clean integration of the platform sections methods.

Note: this is still draft - not ready for merging yet.

### DIFF
--- a/httomo/wrappers_class.py
+++ b/httomo/wrappers_class.py
@@ -1,7 +1,7 @@
-from typing import Dict, Union
+from typing import Dict, Tuple, Union
 import numpy as np
 import inspect
-from inspect import signature
+from inspect import Parameter, signature
 from httomo.utils import print_once
 from httomo.data import mpiutil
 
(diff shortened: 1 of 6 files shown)


httomo/wrappers_class.py
```diff
@@ -287,16 +290,37 @@ def __init__(
             if function_name == "rotation":
                 self.wrapper_method = super()._execute_rotation
 
-        # httomolib can include GPU/CuPy methods as as well as the CPU ones. Here
-        # we check if the method can accept CuPy arrays by looking into docstrings.
+        # httomolib exports metadata from the method decorator, which we can use to
+        # check if we support CuPy
+        func = getattr(self.module, method_name)
+        self.meta = func.meta
```
dkazanc, line 296: what happens if the meta data is not available as with the TomoPy module? Or is it initialised somewhere?

team-gpu: This is part of the `HttomolibWrapper`, so this only happens within the httomolib methods, where the `meta` field is available. But this code isn't used at this point - will integrate that today and see it it can all work together.

PR #129, opened 2023-05-17
Remove tomopy from requirements

jessicavers:


docs/source/doc-conda-requirements.yml
```diff
@@ -11,7 +11,6 @@ dependencies:
   - jinja2
   - sphinx-panels
   - sphinx-copybutton
-  - tomopy
   - pyyaml
   - ipython
```
namannimmo10, line 15: we can also remove ipython?

jessicavers: Keeping in the meantime for use with jupyter notebook

PR #133, opened 2023-05-23
Refactor the task runner

namannimmo10:
- [x] Add a data class for reslicing info
- [x] merge `_initialise_datasets` and `_initialise_stats` into one function
- [x] remove `_check_save_result`
- [x] Add more pipeline integration tests (using `i12`'s data, `!Sweep` functionality, multi inputs)
- [x] Add MPI tests
- [x] Add `LoaderData` and `MethodFunc` dataclasses

### DIFF
--- a/httomo/task_runner.py
+++ b/httomo/task_runner.py
@@ -183,15 +183,7 @@ def run_tasks(
 
     loader_start_time = time.perf_counter_ns()
 
-    (
-        data,
-        flats,
-        darks,
-        angles,
-        angles_total,
-        detector_y,
-        detector_x,
-    ) = _run_loader(method_funcs[0][1], method_funcs[0][3])
+    loader_info = _run_loader(method_funcs[0][1], method_funcs[0][3])
 
     output_str_list = [
         f"    Finished task 1 (pattern={method_funcs[0][1].pattern.name}): {loader_method_name} (",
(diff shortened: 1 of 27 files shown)


tests/test_pipeline.py
```diff
@@ -215,6 +215,51 @@ def test_diad_testing_pipeline_output(
...
+    imarray = np.array(Image.open(tif_files[0]))
+    assert imarray.shape == (128, 160)
+    assert imarray.sum() == 3855857
+
+    imarray = np.array(Image.open(tif_files[1]))
+    assert imarray.shape == (128, 160)
+    assert imarray.sum() == 3856477
+
+    h5_files = list(filter(lambda x: ".h5" in x, files))
+    assert len(h5_files) == 2
```
yousefmoazzam, line 250: Given that this test is for ensuring that parameter sweeps run as expected, and the fact that a parameter sweep on a method produces an intermediate hdf5 file with a different structure compared to normal intermediate hdf5 files, would it be worth:
- checking that the expected datasets exist within the hdf5 file
- similar to the check on the tiff files, do a check of the sum and shape of the datasets in the hdf5 file

namannimmo10: Fair points. I was about to ask about this because when I test the datasets in the hdf5 files, I get an error:

```python
    with h5py.File(h5_files[0], "r") as f:
        print(list(f.keys()))
        print(f["data"].dtype)
```

```console
$ python -m pytest tests/test_pipeline.py::test_sweep_pipeline_with_save_all -rP
                                                                                                         

tests/test_pipeline.py::test_sweep_pipeline_with_save_all FAILED                                                                                            [100%]

============================================================================ FAILURES =============================================================================
________________________________________________________________ test_sweep_pipeline_with_save_all ________________________________________________________________
Traceback (most recent call last):
  File "/scratch/naman/oss/httomo/tests/test_pipeline.py", line 256, in test_sweep_pipeline_with_save_all
    print(f["data"].dtype)
AttributeError: 'Group' object has no attribute 'dtype'
---------------------------------------------------------------------- Captured stdout call -----------------------------------------------------------------------
['data']
```

namannimmo10: Many thanks for documenting this, Yousef.

httomo/task_runner.py
```diff
@@ -216,30 +263,22 @@ def run_tasks(
...
             possible_extra_params,
-            len(method_funcs),
-            func_wrapper,
-            method_funcs[idx][1],
-            method_funcs[idx - 1][1],
+            method_func,
+            method_funcs[idx - 1],
+            method_funcs[idx + 1] if idx < len(method_funcs) - 1 else None,
             dict_datasets_pipeline,
             httomo.globals.run_out_dir,
```
namannimmo10, line 278: no need to have this as an argument to the function.

namannimmo10: thought: see if having a data class `PipelineInfo` would be useful that contains some general info about the pipeline, like `save_all`, `no_of_tasks` etc - info that is shared by all/some of the methods.

namannimmo10: todo: add tests for `MethodFunc`.

PR #136, opened 2023-05-30
CPU-only httomo for CPU tests?

yousefmoazzam:
After dealing with https://github.com/DiamondLightSource/httomo/pull/134 I noticed that it would have been caught much sooner if even a *single* pipeline was executable on CI. This of course isn't possible yet due to requiring GPU-support even when httomo is running a CPU-only pipeline.

I was curious how much effort would be needed to make httomo work with only CPU dependencies, and it didn't seem like too much at a first attempt.

And by the way, with this PR, I'm not suggesting that this should definitely go into `main`; the PR is mainly to trigger the CI.

Feel free to comment if interested!

### DIFF
--- a/httomo/wrappers_class.py
+++ b/httomo/wrappers_class.py
@@ -21,7 +21,14 @@ try:
 except ImportError:
     import numpy as xp
 
-    print("CuPy is not installed")
+    log_once(
+        "CuPy is not installed",
+        # The `comm` parameter for `log_once()` isn't used in the function body,
+        # so any value is fine to pass
+        comm="comm",
+        colour=Colour.LYELLOW,
+        level=1,
+    )
 
 
 def _gpumem_cleanup():
@@ -49,7 +56,13 @@ class BaseWrapper:
             Union[tuple, xp.ndarray, np.ndarray]: transferred datasets
         """
         if not gpu_enabled:
-            return args
+            # Apply the same logic as the `if self.cupyrun` block later on, to
+            # return either a single array or a tuple of arrays based on the
+            # number of args
+            if len(args) == 1:
+                return args[0]
+            else:
+                return args
         xp.cuda.Device(self.gpu_id).use()
         _gpumem_cleanup()
         if self.cupyrun:
@@ -93,7 +106,13 @@ class BaseWrapper:
         data = getattr(self.module, method_name)(data, **dict_params_method)
         if reslice_ahead or save_result and gpu_enabled:
             # reslice ahead, bring data back to numpy array
-            return xp.asnumpy(data)
+            try:
+                return xp.asnumpy(data)
+            except AttributeError:
+                # Getting this error indicates that `xp` is `np`, where `np`
+                # does not have the `asnumpy()` function. This means that `data`
+                # is a numpy array, so return it as is.
+                return data
         else:
             return data
 
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -42,7 +42,6 @@ requires-python = ">=3.9"
 version = "0.0.1"
 dependencies = [
     "click",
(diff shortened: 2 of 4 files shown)


httomo/wrappers_class.py
```diff
@@ -20,8 +20,14 @@
         print("CuPy is installed but GPU device inaccessible")
 except ImportError:
     import numpy as xp
-
-    print("CuPy is not installed")
+    log_once(
+        f"CuPy is not installed",
```
namannimmo10, line 24: No interpolated variables so no f-string needed.
```suggestion
        "CuPy is not installed",
```

pyproject.toml
```diff
@@ -55,6 +54,7 @@ repository = "https://github.com/DiamondLightSource/HTTomo"
 [project.optional-dependencies]
 tomopy = [ "tomopy" ]
 httomolib = [ "httomolib" ]
+cupy = ["cupy"]
```
namannimmo10, line 57: Since only the httomolib backend requires cupy for our project at the moment, should we add cupy dependency to it?

```toml
httomolib = ["httomolib", "cupy"]
```

namannimmo10 (commented):
Great, thank you very much!

I have a few comments.

1. You are making cupy as an optional dependency in httomolib in https://github.com/DiamondLightSource/httomolib/commit/ddd946a9c8a6c7076d363bf1dae9682ce96b6ab3. I am not sure we should do that because httomolib is a package for GPU-methods. We have been getting cupy import errors from files in httomolib while running the CPU tests in Httomo, but you have fixed that in https://github.com/DiamondLightSource/httomolib/commit/59922ef31ef688cd488cbd93404a3a9333ef8c5c.
2. Removing imports in top level `__init__.py` should be fine unless you want to import methods directly like this.
```python
from httomolib import save_to_images
```

We are able to successfully run all the pipeline integration tests with this PR, so I am +1 to merge. Cheers!

yousefmoazzam:
Thanks for the feedback! :)

Yeah, regarding point 1, I wasn't really sure what to do with the production dependencies in `pyproject.toml` to be honest. As far as I am aware, the main way to install httomolib in production will be through `conda` and not `pip`. and we only ever use `pip` to install httomolib in development mode (please correct that if I'm wrong though)?

My logic was that in either case, production GPU or production CPU, all the necessary dependencies are installed via conda's environment YAML file rather than the stuff in `pyproject.toml`, so it didn't really matter to remove CuPy as a production dependency of httomolib in `pyproject.toml`.

Put another way, it seemed to me that the dev dependencies are the only dependencies that matter in `pyproject.toml`, not the production ones, because conda takes care of the production dependencies in both cases:
- production GPU
- production CPU only

(By the same logic though, perhaps I didn't necessarily even need to remove CuPy as a production dependency of httomolib! I did it in a somewhat paranoid attempt to try and be extra sure that CuPy didn't appear in the test env of httomo's CI).

I think perhaps this indicates that there should be some more though going into which files takes care of what dependencies:
- conda's environment YAML
- pip's `pyproject.toml`

because as things currently are, it seems a bit muddled to me.

For example, I personally never install the _production_ dependencies of httomo or httomolib with `pip`, it's always with `conda`. I only ever install _dev_ dependencies with `pip`. What do other people do? @dkazanc @namannimmo10

httomo/wrappers_class.py
yousefmoazzam: Good point, I was being too willing to copy+paste and I missed that :sweat_smile:

yousefmoazzam:
After some discussion we decided to go a different route to enable httomo to run with only CPU dependencies, which will be cleaner in the long run. So, will close this PR and keep it as a reference.

PR #137, opened 2023-05-30
Add CLI flag for naming the output dir created by HTTomo

### LINKED ISSUE #131: Allow users the option to pass the name of the output directory that HTTomo creates
This was a feature available with Savu via the `-f` flag, so it might be useful to add similar functionality in HTTomo's CLI.

One use-case is the autoprocessing done on I18, where the name of the output directory of the Savu run was determined in advance so then the output could be passed onto the next step in the pipeline/zocalo recipe.

yousefmoazzam:
Not much explanation needed I guess, pretty simple.

Fixes #131.

### DIFF
--- a/httomo/cli.py
+++ b/httomo/cli.py
@@ -86,6 +86,12 @@ def check(yaml_config: Path, in_data: Path = None):
     else context.params["out_dir"],
     help="Directory for reslice intermediate files (defaults to out_dir, only relevant if --reslice is also given)",
 )
+@click.option(
+    "--output-folder",
+    type=click.Path(exists=False, file_okay=False, writable=True, path_type=Path),
+    default=None,
+    help="Optionally define the name of the output folder created by HTTomo",
+)
 def run(
     in_file: Path,
     yaml_config: Path,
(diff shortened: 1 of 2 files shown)


namannimmo10 (approved):
Thanks, LGTM. Do you want to add a test for this in `test_cli.py`? A simple test that checks if the user log path is updated.

yousefmoazzam:
Good idea, I'll add a test!

yousefmoazzam:
Re the linter failing: dunno how others feel, I tend to dislike including changes in a PR that are unrelated to the intended feature, it feels a bit weird to include unrelated edits to `task_runner.py` in a PR for adding a flag to the CLI?

However, if people are ok with applying `black` to fix those errors in this PR, and it's preferable over fixing the linter errors separately, then I'm fine with doing that here, no biggie.

namannimmo10:
we can automate it if you want: https://github.com/DiamondLightSource/httomo/commits?author=lint-action

yousefmoazzam:
Mmm, that's one option. Though it results in those commits by bots which are a bit useless.

I wonder what other groups/teams at DLS do regarding enforcement of code formatting, because I don't think I've see those bot linter commits in Controls Group stuff for example, and they have a strict adherence to `black` code formatting I think? :thinking: They probably use a vscode pluign that automatically applies `black` whenever a python file is saved come to think of it, I remember seeing that somewhere, which is perhaps a bit too intrusive.

By the way, absolutely feel free to say if it's a bit too pedantic to care much about, that's an understandable answer to stuff like this! Once I discovered squashing with interactive rebasing, I now know I can retroactively apply `black` and never have to make those useless commits that just apply the code formatter. It's enabling me too much now in my hope to keep my commit history tidy haha.

PR #138, opened 2023-06-05
update httomo with the latest changes in httomolib and httomolibgpu

namannimmo10:


httomo/wrappers_class.py
```diff
@@ -302,7 +354,7 @@ def __init__(
             if function_name == "rotation":
                 self.wrapper_method = super()._execute_rotation
 
-        # httomolib exports metadata from the method decorator, which we can use to
+        # httomolibgpu exports metadata from the method decorator, which we can use to
         # check if we support CuPy
         func = getattr(self.module, method_name)
         self.meta = func.meta
```
namannimmo10, line 360: since `httomolibgpu` is GPU-only, `self.cupyrun` can be `True` here?

dkazanc: yes! in fact the info from decorators (cpu or gpu) becomes also a bit redundant but we can deal with it later. We can now do checks just based on the package name `httomolib` and `tomopy` both CPU and `httomolibgpu` is GPU

PR #146, opened 2023-06-21
Specify tiff and h5 files in output dir to load in tests

yousefmoazzam:
We saw some potential ambiguity in the ordering of files when running some tests (due to different workstations producing failing or passing tests), such as in https://github.com/DiamondLightSource/httomo/blob/memory/tests/test_pipeline.py#L444-L455.

In this PR, the tests which depended on using indices to determine which files to load have been changed to specify a substring of the desired filename instead, in an attempt to avoid ambiguity.

### DIFF
--- a/tests/test_pipeline.py
+++ b/tests/test_pipeline.py
@@ -441,15 +441,25 @@ def test_multi_inputs_pipeline(cmd, standard_data, sample_pipelines, output_fold
     h5_files = list(filter(lambda x: ".h5" in x, files))
     assert len(h5_files) == 3
 
-    with h5py.File(h5_files[0], "r") as f:
+    median_filter_tomo = list(
+        filter(lambda x: "median_filter3d-tomo.h5" in x, h5_files)
+    )[0]
+    median_filter_flats = list(
+        filter(lambda x: "median_filter3d-flats.h5" in x, h5_files)
+    )[0]
+    median_filter_darks = list(
+        filter(lambda x: "median_filter3d-darks.h5" in x, h5_files)
+    )[0]
+
+    with h5py.File(median_filter_flats, "r") as f:
         arr = np.array(f["data"])
         assert arr.shape == (20, 128, 160)
         assert arr.dtype == np.uint16
-    with h5py.File(h5_files[1], "r") as f:
+    with h5py.File(median_filter_darks, "r") as f:
         arr = np.array(f["data"])
         assert arr.shape == (20, 128, 160)
         assert arr.dtype == np.uint16
-    with h5py.File(h5_files[2], "r") as f:
+    with h5py.File(median_filter_tomo, "r") as f:
         arr = np.array(f["data"])
         assert arr.shape == (180, 128, 160)
         assert arr.dtype == np.uint16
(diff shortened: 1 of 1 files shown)


tests/test_pipeline.py
```diff
@@ -381,8 +434,10 @@ def test_sweep_pipeline_with_save_all_using_mpi(
...
 
-    log_files = list(filter(lambda x: ".log" in x, files))
-    assert len(log_files) == 2
+    serial_log_files = list(filter(lambda x: ".log" in x, serial_files))
+    parallel_log_files = list(filter(lambda x: ".log" in x, parallel_files))
+    assert len(serial_log_files) == 1
+    assert len(parallel_log_files) == 1
 
 
 """
```
namannimmo10, line 443: Thanks for the PR. I don't think we should comment out tests.

tests/test_pipeline.py
```diff
@@ -204,28 +204,36 @@ def test_i12_testing_pipeline_output(
...
-        total_sum += arr.sum()
+    # total_sum = 0
+    # for i in range(10):
+    #    arr = np.array(Image.open(tif_files[i]))
+    #    assert arr.dtype == np.uint8
+    #    assert arr.shape == (192, 192)
+    #    total_sum += arr.sum()
 
-    assert total_sum == 25834244.0
+    # assert total_sum == 25834244.0
```
yousefmoazzam, line 214: I've commented out the assertion on the sum over all tiff files, since Daniil and I found 3 different cases where that the sum produces 3 different results (me running the test on my machine `pc0074`,  me running the test on `ws381`, and Daniil running the test on his machine). It seems weird, and the rest of the test is still testing the h5 files created, so at least something is being done to test the result of this pipeline?

An alternative would be to provide a tolerance to that assertion. One value that was seen was `25834156.0` , which is in the order of the 10's that it's off compared to the original value, and I suppose there could be other cases where the sum is further away from the original value (we won't know for sure unless we find the cause I guess), so it's hard to know what is a reasonable tolerance. 

Thoughts? I know previously @namannimmo10 mentioned concern about commenting out tests.

tests/test_pipeline.py
namannimmo10: fixed!

tests/test_pipeline.py
```diff
@@ -431,12 +439,9 @@ def test_sweep_pipeline_with_save_all_using_mpi(
...
     assert len(parallel_log_files) == 1
 
+    log_contents = _get_log_contents(serial_log_files[0])
+    mpi_log_contents = _get_log_contents(parallel_log_files[0])
 
-"""
-#   Something weird going on here with the logs 
-
-    mpi_log_contents = _get_log_contents(log_files[1])
-    log_contents = _get_log_contents(log_files[0])
```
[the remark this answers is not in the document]
namannimmo10: Well, my bad, should've added a new commit here or modified the existing one - didn't read the commit message. Usually, people [write good commit messages](https://cbea.ms/git-commit/) with the relevant changes, but we've not been strict about it. I'd simply "squash and merge" from the button down below and write one good commit message.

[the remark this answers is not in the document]
yousefmoazzam: No worries, it's a benign thing really, but since it's such a simple case it's also a good example for learning!

Thanks for the link on writing good commit messages, I tend to use commit messages as places to put useful info in case I were to refer back to them in the future, but I'm sure I could improve mine :muscle:

PR #148, opened 2023-06-23
Version from tag

yousefmoazzam:
This has the same intent as this PR for httomolib https://github.com/DiamondLightSource/httomolib/pull/1


yousefmoazzam:
FYI, the stuff to make the CI work with "CPU-only httomo" has not yet been done, so this stuff will fail to run on the CI as it currently is.

I'll make another PR to be merged into this one that'll make the necessary changes to get "CPU-only httomo" to run. It'll be similar in approach to the other PR that we did a bit ago (https://github.com/DiamondLightSource/httomo/pull/136) but chose to close as we decided instead to split the methods repo into httomolib and httomolibgpu to make things a bit nicer.

.github/workflows/nightly_build.yml
```diff
@@ -0,0 +1,40 @@
...
+
+      - name: Decrypt a secret
+        run: ./.scripts/decrypt_secret.sh
+        env:
+          LARGE_SECRET_PASSPHRASE: ${{ secrets.LARGE_SECRET_PASSPHRASE }}
+
+      - name: Build and upload the package to httomo conda cloud
+        env:
+          LABEL: dev
+        run: |
```
namannimmo10, line 37: Shall we run the tests before uploading it to the conda cloud? It runs at midnight, so I think we should avoid pushing failing code to the channel (if in case the CI is failing, for instance now).

yousefmoazzam: As far as I understood `conda build`, the `meta.yaml` allows the user to specify which tests are run during the "test phase" of `conda build` https://github.com/DiamondLightSource/httomo/blob/version-from-tag/conda/recipe/meta.yaml#L30. So when the `conda_upload.sh` script is executed, on this line https://github.com/DiamondLightSource/httomo/blob/version-from-tag/.scripts/conda_upload.sh#L22 it will run those tests as part of the build process. If any of these tests fail, the build process fails, and the package does not have a chance to get uploaded to the cloud.

For example, I think this is the reason why the upload script for httomolibgpu has the `--no-tests` flag https://github.com/DiamondLightSource/httomolibgpu/blob/main/.scripts/conda_upload.sh#L22; otherwise, the tests would run in the CI and would fail due to lack of a GPU and we would never be able to upload the httomolibgpu package to the conda cloud.

This is how things appear to work when using `conda build` locally on the command line too; if the build fails during the tests, then the built package gets moved to some directory named `broken/` and it will not print out information about how to upload the built package to the cloud.

So I was thinking that we could choose which tests to run (ie, CPU-only when CPU-only httomo is ready) during the build process via the `meta.yaml`, which then governs whether the package (dev or prod) should be uploaded to the cloud. This also then allows us to run the relevant tests only _once_ during the CI run (ie, during the build process), rather than twice:
- once, separate from the build
- another time, during the build

I could be wrong about how `conda build` works regarding running tests though, so please do check if in doubt!

namannimmo10: It's good that we have the "test" phase of conda build in `meta.yml`, which indeed runs `pytest` to execute the test suite. We can separately run CPU tests on the CI and then upload httomo to the conda cloud (without the "test" phase in `meta.yml`.)

If you look at all the [feedstocks available on conda forge](https://github.com/conda-forge/feedstocks/tree/main/feedstocks) and open up the meta yamls, you'd see that many repos do not execute their test suites (with `command`) in the "test" phase... they only do `imports`.

namannimmo10: It is your own choice, really. :)

yousefmoazzam: That's a fair point, we could just have `import` tests in the build process, and have the main test suite run separately.

However, some things regarding the installation of the package are missed with doing purely `import` statements, such as what I saw with YAML files not being copied correctly during installation - the error was only detectable via installing+running in a new env (which _does_ occur during the "test phase" of the build process if the test suite is run, but not if only `import` tests are done).

Since the installation of httomo could well change, until we have stability across httomo, httomolib, and httomolibgpu, I think I'd prefer to have the test suite run during the "test phase" of the build process, to catch any potential issues with the installation of httomo that may otherwise be missed.

When there's some stability of the installation of all these packages, then I think it'd be fine to transition to what you've suggested/what most others are doing.

namannimmo10: good idea, thanks!

PR #149, opened 2023-06-23
CPU-only HTTomo for running CPU tests on CI

yousefmoazzam:
An improvement upon https://github.com/DiamondLightSource/httomo/pull/136, after the decision to split the repo of methods into httomolib (CPU) and httomolibgpu (GPU).

### DIFF
--- a/httomo/cupy_utils.py
+++ b/httomo/cupy_utils.py
@@ -25,3 +25,19 @@ except ImportError:
         colour=Colour.LYELLOW,
         level=1,
     )
+
+
+def _get_available_gpu_memory(safety_margin_percent: float = 10.0) -> int:
+    try:
+        import cupy as cp
+
+        dev = cp.cuda.Device()
+        # first, let's make some space
+        pool = cp.get_default_memory_pool()
+        pool.free_all_blocks()
+        cache = cp.fft.config.get_plan_cache()
+        cache.clear()
+        available_memory = dev.mem_info[0] + pool.free_bytes()
+        return int(available_memory * (1 - safety_margin_percent / 100.0))
+    except:
+        return int(100e9)  # arbitrarily high number - only used if GPU isn't available
--- a/httomo/task_runner.py
+++ b/httomo/task_runner.py
@@ -33,6 +33,7 @@ from httomo.utils import (
     log_once,
     remove_ansi_escape_sequences,
 )
+from httomo.cupy_utils import _get_available_gpu_memory
 from httomo.wrappers_class import HttomolibWrapper, HttomolibgpuWrapper, TomoPyWrapper
 from httomo.yaml_utils import open_yaml_config
 
@@ -736,22 +737,6 @@ def _determine_platform_sections(
     )
 
     return ret
-
-
-def _get_available_gpu_memory(safety_margin_percent: float = 10.0) -> int:
-    try:
-        import cupy as cp
-
-        dev = cp.cuda.Device()
-        # first, let's make some space
-        pool = cp.get_default_memory_pool()
-        pool.free_all_blocks()
-        cache = cp.fft.config.get_plan_cache()
-        cache.clear()
-        available_memory = dev.mem_info[0] + pool.free_bytes()
-        return int(available_memory * (1 - safety_margin_percent / 100.0))
-    except:
-        return int(100e9)  # arbitrarily high number - only used if GPU isn't available
 
 
 def _update_max_slices(


yousefmoazzam:
The branch to merge into has temporarily been set to `main` rather than `version-from-tag` just to be able to trigger the CI easily; the plan is still to merge into `version-from-tag`.

namannimmo10:
if upgrading the `numexpr` version doesn't solve the problem, you can filter `DeprecationWarning` in `pyproject.toml`

yousefmoazzam:
Aha, I wasn't aware of this very useful feature of `pytest` to filter warnings during running tests https://docs.pytest.org/en/7.3.x/how-to/capture-warnings.html, thanks @namannimmo10! :slightly_smiling_face:

yousefmoazzam:
Ok, two things that are worth bringing to people's attention:
- 4c44dd5 does something questionable, so please read its commit message and decide if it's too hacky or not...
- 9af586c is a bit of a workaround to not knowing how to nicely figure out what is wrong with the installation of tomopy to cause the error cited in the commit message
  - The test in question passes fine when pytest is told to ignoring the fact that some optional library for accelerating methods is not available during the CI run (an `ImportWarning` is raised)
  - It's likely not a big deal to do this temporarily, but perhaps people would prefer to fix it properly by investigating the tomopy installation in the CI run, so please say so if someone would prefer to do that than what I've done as a workaround

namannimmo10:
You can move `xp` to a separate module like `cupy_utils.py` and all the files import from it. That module may also include stuff like https://github.com/DiamondLightSource/httomo/blob/1fcacef7b1ef7572aff800eb9d0f132aebe682df/httomo/task_runner.py#L741

yousefmoazzam:
Sure, feel free to try that approach and see if it works. I tried moving `xp` to both `globals.py` and `cli.py` (in an attempt to assign `xp` its proper value ASAP after the run was initiated via the CLI) without much success, but perhaps there is a place where it can be moved that works.

yousefmoazzam:
I'll try creating a `cupy_utils.py` as suggested and I'll see what I can do with it.

yousefmoazzam:
Cool, it seems like creating `cupy_utils.py` and moving the definition of `xp` to it worked! Thanks for the suggestion @namannimmo10  :slightly_smiling_face: 

I think the way I was trying to do it where I:
- have `xp` initialised as `None` in `globals.py`
- assign `xp` its proper value near the start of the program launching (for example, in `cli.py`)

was too strict or something, it seemed difficult to have a central place with _all_ global variables that get assigned their proper values at some point during execution of the program.

Initially I was sceptical of the approach of defining module-level variables in the manner that `xp` now is, because I thought it was too erratic that the module-level variable would be defined only once the module has been imported (primarily because the order of imports can change easily without much thought).

But after thinking about it, perhaps it's fine actually, because _whenever_ the module-level variable is needed it's going to be imported, and so the module-level variable gets defined as soon as it's being asked for by the `import`.

Having a specific place in the code where the global variable is assigned its proper value would be nice (ie, the way I originally tried), but I guess it's still pretty decent that you can point to the _earliest_ place in program execution where the module with the global variable is imported, and _that_ would be when it gets assigned its proper value.

namannimmo10:
Is this ready to go in?

pyproject.toml
```diff
@@ -53,10 +54,21 @@ httomo = "httomo.cli:main"
 repository = "https://github.com/DiamondLightSource/HTTomo"
 
 [project.optional-dependencies]
-tomopy = [ "tomopy" ]
-httomolib = [ "httomolib" ]
-httomolibgpu = ["httomolibgpu", "cupy"]
-dev = [
+cpu = ["tomopy" ]
+gpu = ["httomolibgpu"]
```
namannimmo10, line 58:
```suggestion
gpu = ["httomolibgpu", "cupy"]
```

yousefmoazzam: I think httomolibgpu already specifies cupy as a dependency https://github.com/DiamondLightSource/httomolibgpu/blob/main/pyproject.toml#L41, so I didn't know if this was necessary or not; do you know if `pip` in combination with dependencies in `pyproject.toml` figures out "dependencies of dependencies" nicely?

namannimmo10: yep, right, it should be in the environment already then.

yousefmoazzam:
I'm generally happy with this as it is, dunno if @dkazanc wanted to take a look before it goes in?

namannimmo10:
It looks good to me, thanks. I am skeptical about the nightly build though, there isn't a need for that. It is possible for this to upload code with failing tests to the conda cloud since we don't always make sure the `main` branch is up and running.

BTW, CI log:

> ```
> Packaging httomolibgpu
> INFO:conda_build.build:Packaging httomolibgpu-1.0-py310_0
> Packaging httomolibgpu-1.0-py310_0
> Importing conda-verify failed.  Please be sure to test your packages.  conda install conda-verify to make this message go away.
> WARNING:conda_build.build:Importing conda-verify failed.  Please be sure to test your packages.  conda install conda-verify to make this message go away.
> ```

yousefmoazzam:
I don't think it's possible for the nightly build to upload to conda cloud if the tests fail, my limited experience with `conda build` is that if the tests fail during the build, then the built file (`.tar.gz2` or something) gets moved to some `broken/` dir (rather than a dir named as the architecture that the package was built for) and nothing is done with it, the `conda build` command technically fails and so I don't think it proceeds with the upload?

However, even if the built package w/ failing tests on `main` _does_ get uploaded to conda cloud, it'll get uploaded to the `dev` "label" in the httomo channel. The default label is `main`, and is where packages are installed from if no label is specified, according to the conda/anaconda.org docs https://docs.anaconda.com/reference/glossary/#glossary-label.

In order for someone to install the nightly build, they would have to specify the `dev` label like the following
```
conda install -c httomo/label/dev httomo
```

If someone does simply
```
conda install -c httomo httomo
```
then they will never be installing something from the `dev` label.

That way, only developers would really be installing the nightly builds, and I don't think developers would really mind code with failing tests? When Daniil and I chatted about it, it seemed like what we cared more about with these nightly builds was having a snapshot of what the package looked like (not just the code, but the entire built package) on the `main` branch.

Dunno if that addresses your concerns though?

And regarding the `conda-verify` warning, we're aware of that; I don't tend to worry about that warning since I've not seen anything go wrong if I _don't_ use `conda-verify`, but I agree that maybe making it go away could be easy enough :sweat_smile: . Seems like it's not been updated in a while, last release was in 2019 https://github.com/conda/conda-verify/releases, but maybe it still does something useful.

namannimmo10 (approved):
I see, thanks for the info. I'm +1 to merge then.

PR #154, opened 2023-06-29
Move reslice to section loop

yousefmoazzam:
This has moved the reslice operation from the `run_method()` function to the sections-loop.

Note that the execution of methods is still in `run_method()`, so the current state is the following:
- reslices are being done within the sections-loop
- the sections-loop is not executing any methods yet
- _after_ the sections-loop finishes, methods are executed via a loop calling `run_method()`

Therefore, the reslices are **not** happening in the correct places in the execution of the pipeline. Once the execution of the methods has been moved to the methods-loop (the innermost loop within the sections-loop), this should be resolved.

### DIFF
--- a/httomo/task_runner.py
+++ b/httomo/task_runner.py
@@ -171,7 +171,6 @@ def run_tasks(
         (["comm"], comm),
         (["out_dir"], httomo.globals.run_out_dir),
         (["save_result"], False),
-        (["reslice_ahead"], False),
     ]
     # data shape and dtype are useful when calculating max slices
     data_shape = loader_info.data.shape
@@ -246,6 +245,15 @@ def run_tasks(
         # Assume that, after every section has been completed (except for the
         # last section), a reslice should occur
         if i < len(platform_sections) - 1:
+            reslice_info.count += 1
+            if reslice_info.count > 1 and not reslice_info.has_warn_printed:
+                reslice_warn_str = (
+                    "WARNING: Reslicing is performed more than once in this "
+                    "pipeline, is there a need for this?"
+                )
+                log_once(reslice_warn_str, comm=comm, colour=Colour.RED)
+                reslice_info.has_warn_printed = True
+
             next_section_in = platform_sections[i+1].methods[0].parameters["data_in"]
             # TODO: should `slicing_dim_section` be used to define
             # `current_slice_dim`? Or would that be too confusing to randomly
@@ -292,7 +300,7 @@ def run_tasks(
             method_func.parameters.update({"ncore": ncore})
 
         idx += 1
-        reslice_info, glob_stats[idx] = run_method(
+        glob_stats[idx] = run_method(
             idx,
             save_all,
             possible_extra_params,
@@ -302,7 +310,6 @@ def run_tasks(
             dict_datasets_pipeline,
             glob_stats[idx],
             comm,
-            reslice_info,
         )
 
         stop = time.perf_counter_ns()
@@ -423,7 +430,6 @@ def _get_method_funcs(yaml_config: Path, comm: MPI.Comm) -> List[MethodFunc]:
             is_loader=True,
             cpu=True,
             gpu=False,
-            reslice_ahead=False,
             pattern=Pattern.all,
         )
     )
@@ -466,7 +472,6 @@ def _get_method_funcs(yaml_config: Path, comm: MPI.Comm) -> List[MethodFunc]:
                 calc_max_slices=None
                 if not is_httomolibgpu
                 else wrapper_init_module.calc_max_slices,
-                reslice_ahead=False,
                 pattern=Pattern.all,
                 is_loader=False,
                 is_last_method=True if i == methods_count - 2 else False,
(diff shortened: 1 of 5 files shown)


httomo/task_runner.py
```diff
@@ -266,17 +300,14 @@ def run_tasks(
             method_func.parameters.update({"ncore": ncore})
 
         idx += 1
-        reslice_info, glob_stats[idx] = run_method(
+        glob_stats[idx] = run_method(
```
namannimmo10, line 303: we don't really need `glob_stats` that we get from `run_method`, do we?. So, run method can just _run_ method, not return anything.

yousefmoazzam: It's a good question, I'm not sure to be honest.

I think stats from each method can be useful info to provide at the end of the run or something, which is maybe why Daniil asked me to put `glob_stats` in. When he's back we can check if we need the stats info from a method execution to be kept afterwards or not!

httomo/task_runner.py
```diff
@@ -243,6 +241,42 @@ def run_tasks(
...
+            else:
+                resliced_data, _ = reslice_filebased(
+                    dict_datasets_pipeline[next_section_in],
+                    current_slice_dim,
+                    next_slice_dim,
+                    comm,
+                    reslice_dir,
+                )
+            # Store the resliced data
+            dict_datasets_pipeline[next_section_in] = resliced_data
```
yousefmoazzam, line 279: Can probably move some of the logic to a separate function `_perform_reslice()` or something, to make the sections-loop a bit nicer, I'll see how such a function ends up looking

httomo/task_runner.py
dkazanc: I think by default `glob_stats` are associated with every method result as other methods in the pipeline might request it. However, with the blocks loop this makes it a bit tricky...

It seems that the stats should be calculated after the loop over blocks is finished, so only in the sections loop, where the data on the cpu is available as a whole. 

This is not ideal of course... As for global_stats in image saver we should be probably OK as it is the CPU section, but for stats that are needed, say, in the median filter we cannot much to offer except `local_stats` which is not a great idea as the data in blocks will be treated differently. So I think we should be aware of that... 

Could we for now remove `glob_stats` from `run_method` but rather add it to the end of the blocks loop?

yousefmoazzam: Yeah, I see your point that the blocks loop kinda gets in the way of calculating stats on the result of each method.

Sure, I'm happy to make that change. Just to make sure I've understood this correctly, does this mean that stats will only be calculated on the results of methods that are at _the end of sections_, since we'd be waiting for all blocks to complete in a section before doing stats stuff?

dkazanc: yes, it seems so that stats will be **only** available for the last method in the section. 

May be there is more torturous way to make it work for every method by calculating stats on cupy arrays for every block and then collect it afterwards, but for now, I don't think that we should do it. Crucially we need stats while saving images and this is a new section, so we can get it calculated for it. 

So I guess we should change the terminology to `global_stats_section` or something? And may be it will more stored in the meta data of the section?

yousefmoazzam: Hmm, let's see... Currently, `glob_stats` is a dict associating stats with dataset names. If we want the option in the future  to associate stats to the result of every method, then it's probably best to keep that with the dict structure it currently has, and not rename it to `glob_stats_section`?

For the stats of a section, _this_ would make sense to be in the `PlatformSection` dataclass, we can call it something like `output_stats` or something, and then we use this as the way to easily pass stats into the image saver for now?

dkazanc: sounds good, thanks.

httomo/task_runner.py
```diff
@@ -243,6 +245,28 @@ def run_tasks(
...
+        #
+        # NOTE: Currently assuming that the dataset to calculate stats from is
+        # the same input dataset from the beginning of the sections loop; should
+        # this instead be getting the output dataset from the last method that
+        # was in the section which has just completed?
+        section.output_stats = min_max_mean_std(data_full_section, comm)
+
+        # Assume that, after every section has been completed (except for the
+        # last section), a reslice should occur
+        if i < len(platform_sections) - 1:
```
dkazanc, line 259: I guess we currently assume that the reslice happens always at the end of each section? In the pipeline bellow 2 reslices happened while only 1 should be. I guess `should_reslice` previously controlled it? What do we have now instead to check for this?

```
Section 0
httomolibgpu.misc.corr
httomolibgpu.misc.corr
httomolibgpu.misc.corr
httomolibgpu.prep.normalize
Section 1
httomolibgpu.recon.rotation
httomolibgpu.prep.stripe
httomolibgpu.recon.algorithm
WARNING: Reslicing is performed more than once in this pipeline, is there a need for this?
Section 2
httomolib.misc.images
```

dkazanc: May be slightly re-think the platform section stuff: 

- I feel that we need to reslice ONLY when we have a change of the pattern and do not reslice when the architecture changes (and the pattern remains). Would the section need to change in this case? Probably, yes.
- However for the CPU section we don't need to iterate over blocks, and methods can be decoupled and the results saved. _Should every CPU method become a separate section then?_ I know it is a bit drastic, but nothing really holds them together, they are separate entities memory-wise apart from the notion of pattern. However if we keep them in one section then the saving of the data become a bit messy... 
- To conclude may be we need to apply sectioning logic to GPU device-to-device only?

PR #159, opened 2023-07-21
Fix blocks loop

yousefmoazzam:
This is a minimal set of changes to allow the blocks-loop to complete for all sections when running the following YAML on I12 data:
```yaml
- httomo.data.hdf.loaders:
    standard_tomo:
      name: tomo
      dimension: 1
      preview:
        - 
        - start: 1000
          stop: 1200
        - 
      pad: 0
- httomolibgpu.prep.normalize:
    normalize:
      data_in: tomo
      data_out: tomo
      cutoff: 10.0
      minus_log: true
      nonnegativity: false
      remove_nans: false
- httomolibgpu.misc.corr:
    median_filter3d:
      data_in: tomo
      data_out: tomo
      kernel_size: 3
- httomolibgpu.recon.algorithm:
    FBP:
      data_in: tomo
      data_out: tomo
      center: 1290
      objsize: null
```

Summary of changes:
- Run the `prerun_method()` function for each method once per-section, rather than once per-block
- Reuse a `RunMethodInfo` object for a method across different blocks being processed
- Store the processed blocks outputted by a section by overwriting data in the input array
- Create a separate numpy array for the output of a section that contains a recon method, because recon methods change the shape of the data


dkazanc:
So, I've added the centering in the loop, but I cannot say that it is very elegant as there are code [duplications](https://github.com/DiamondLightSource/httomo/blob/d4a3bf10332491a912789293aa10f2806c4904b3/httomo/task_runner.py#L263) and a division [created](https://github.com/DiamondLightSource/httomo/blob/d4a3bf10332491a912789293aa10f2806c4904b3/httomo/task_runner.py#L289) between `res` and `res_param`. The former is the output of the wrapper with the **same** shape and the latter is usually a scalar or a list of scalars - parameters. 
Centering probably goes against our idea of overriding input with output so we need to add some workarounds. So I effectively "silencing" the data copy for centering and after the wrapper done initialising `res` (for first method only). 

It works. I also tried with the stripe removal before centering.

Also the calculation of [stats](https://github.com/DiamondLightSource/httomo/blob/d4a3bf10332491a912789293aa10f2806c4904b3/httomo/task_runner.py#L374) does something naughty and crashes IDE's and terminals. It needs looking into. Also we should enable data saving option to see that we're actually doing the right thing :)

httomo/postrun.py
```diff
@@ -20,32 +20,32 @@
 def postrun_method(
     run_method_info: RunMethodInfo,
     dict_datasets_pipeline: Dict[str, Optional[np.ndarray]],
-    current_func: MethodFunc,
+    section: PlatformSection,
 ):
-    # TODO: What happens when a method with multiple outputs has
-    # `run_method_info.save_result = True`? Should all output datasets be saved?
-    is_3d = len(dict_datasets_pipeline[run_method_info.data_out].shape) == 3
+    if run_method_info.method_name is not "save_to_images" and 'center' not in run_method_info.method_name:
```
namannimmo10, line 25:
```suggestion
    if run_method_info.method_name != "save_to_images" and 'center' not in run_method_info.method_name:
```

dkazanc:
image

PR #162, opened 2023-07-26
Add loader params for ignoring darks/flats

### LINKED ISSUE #157: option to ignore certain flats/darks
we need to add an option to ignore certain flats and/or darks in the loader as sometimes some of them are not usable. 

Ideally we would like to combine a singular selection and a batch selection, e.g. in the list:
`ignore_flats: [5, 8:15, 19]`
`ignore_darks: [0:20]`  

Also by default it can be:
`ignore_flats: False`
`ignore_darks: False`

but when 
`ignore_flats: True`
`ignore_darks: True`
ALL flats and darks must be ignored.... 

Daniil wrote:

> you can test it on Diad's data that they just collected, e.g. sample 23565
> you will see that the normalisation result will have no Infs if you exclude some darks (which are not darks)

> ```
> - httomo.data.hdf.loaders:
>     standard_tomo:
>       name: tomo
>       data_path: /entry/imaging/data
>       image_key_path: /entry/instrument/imaging/image_key
>       rotation_angles:
>         data_path: /entry/imaging_sum/gts_cs_theta
>       dimension: 1
>       preview:
>         - 
>         - start: 1000
>           stop: 1005
>         - start: 700
>           stop: 2000
>       pad: 0
> - httomolibgpu.prep.normalize:
>     normalize:
>       data_in: tomo
>       data_out: tomo
>       cutoff: 10.0
>       minus_log: true
>       nonnegativity: false 
> - httomolibgpu.recon.algorithm:
>     FBP:
>       data_in: tomo
>       data_out: tomo
>       center: 700
>       objsize: null      
> - httomolib.misc.images:
>     save_to_images:
>       data_in: tomo
>       subfolder_name: images
>       axis: 0
>       file_format: tif
>       bits: 8
>       perc_range_min: 0.0
>       perc_range_max: 100.0
>       jpeg_quality: 95
> ```

yousefmoazzam:
Fixes #157

Specifying individual indices of darks/flats is supported, as well as specifying a range of indices via start and stop values.

TODO list:
- [x] Add parameters for ignoring darks/flats
- [x] Add docs for the new loader parameters
- ~OPTIONAL: Move the parameters `ignore_darks` and `ignore_flats` to be within the existing `darks` and `flats` parameters~
- [x] Add pipeline test that ignores some darks and flats, ~which in turn influences the result of the processing (and thus can be verified numerically in the intermediate files)~
- [x] Add basic error checking that the provided darks/flats indices to ignore in the YAML are indeed referring to darks/flats
- [x] Add loader test to verify that the number of returned darks/flats reflects the number that should have been skipped

### DIFF
--- a/docs/source/reference/loaders.rst
+++ b/docs/source/reference/loaders.rst
@@ -1,6 +1,117 @@
 .. _reference_loaders:
 
 HTTomo Loaders
---------------------------
+--------------
 
-TODO.
+Ignoring darks and/or flats
+===========================
+
+There are situations where certain dark or flat fields that have been collected
+in the data need to be excluded from processing. The standard tomography loader
+in HTTomo offers the functionality to exclude darks and/or flat field images
+based on their index in the dataset they are contained in. The images can be
+excluded by specifying either:
+
+- Individual indices
+- A range of indices with start and stop values
+
+This functionality is used in a process list via the :code:`ignore_darks` and
+:code:`ignore_flats` parameters.
+
+To specify individual indices to exclude, use the :code:`individual` subfield,
+and to specify a range of indices to exclude, use the :code:`batch` subfield.
+These subfields can both be used at the same time to select batches of
+darks/flats as well as individual darks/flats to ignore.
+
+It is also possible to ignore *all* darks and/or flats by setting the value of
+either :code:`ignore_darks` or :code:`ignore_flats` to :code:`true`.
+
+Below shows some example uses of the :code:`ignore_darks` parameter, and the
+:code:`ignore_flats` parameter can be used in a similar manner.
+
+Ignore individual darks only
+++++++++++++++++++++++++++++
+
+.. code-block:: yaml
+  :emphasize-lines: 12-13
+
+    - httomo.data.hdf.loaders:
+        standard_tomo:
+        name: tomo
+        data_path: entry1/tomo_entry/data/data
+        image_key_path: entry1/tomo_entry/instrument/detector/image_key
+        dimension: 1
+        preview:
+            -
+            - start: 30
+              stop: 60
+            -
+        ignore_darks:
+            individual: [0, 5]
+
+Ignore all darks
+++++++++++++++++
+
+.. code-block:: yaml
+  :emphasize-lines: 12
+
+    - httomo.data.hdf.loaders:
+        standard_tomo:
+        name: tomo
+        data_path: entry1/tomo_entry/data/data
+        image_key_path: entry1/tomo_entry/instrument/detector/image_key
+        dimension: 1
+        preview:
+            -
+            - start: 30
+              stop: 60
+            -
+        ignore_darks: true
+
+Ignore individual darks and a group of darks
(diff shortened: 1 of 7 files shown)


yousefmoazzam:
Some more info on the optional third task: there already exists the `darks` and `flats` parameters in the loader, which was originally used for specifying when the darks/flats are in separate files: https://github.com/DiamondLightSource/httomo/blob/51470be0feb9727374d9f80d4d10453b47499a56/samples/pipeline_template_examples/DLS/03_i12_separate_darks_flats.yaml#L13-L18

The `ignore_darks` and `ignore_flats` parameters are currently separate from the `darks` and `flats` parameters, and would be used like the following when specifying separate files with darks and flats:
```yaml
- httomo.data.hdf.loaders:
    standard_tomo:
      name: tomo
      data_path: /1-TempPlugin-tomo/data
      rotation_angles:
        user_defined:
          start_angle: 0
          stop_angle: 180
          angles_total: 724
      dimension: 1
      preview: [null, null, null]
      pad: 0
      darks:
        file: tests/test_data/i12/separate_flats_darks/dark_field.h5
        data_path: /1-NoProcessPlugin-tomo/data
      ignore_darks: ...
      flats:
        file: tests/test_data/i12/separate_flats_darks/flat_field.h5
        data_path: /1-NoProcessPlugin-tomo/data
      ignore_flats: ...
```

Maybe it's preferable to move `ignore_darks` and `ignore_flats` into `darks` and `flats` respectively, like so:
```yaml
- httomo.data.hdf.loaders:
    standard_tomo:
      name: tomo
      data_path: /1-TempPlugin-tomo/data
      rotation_angles:
        user_defined:
          start_angle: 0
          stop_angle: 180
          angles_total: 724
      dimension: 1
      preview: [null, null, null]
      pad: 0
      darks:
        file: tests/test_data/i12/separate_flats_darks/dark_field.h5
        data_path: /1-NoProcessPlugin-tomo/data
        ignore: ...
      flats:
        file: tests/test_data/i12/separate_flats_darks/flat_field.h5
        data_path: /1-NoProcessPlugin-tomo/data
        ignore: ...
```

yousefmoazzam:
The docs for these parameters have been placed in a page that was empty and reserved for information about loaders. Feel free to say if people think the docs could be moved somewhere else!

yousefmoazzam:
There was at least 1 difference in the sums/means being calculated on the intermediate datasets saved from the pipeline when comparing:
- results from a run on local workstation
- results from a run in CI

The differences weren't significant (~0.001), but were larger than the tolerance (1e-6) so this difference caused the test running on CI to fail.

The versions of a few packages that could influence the numerical results such as tomopy, scipy, and blas were checked in the CI against the local workstation, but nothing revealed the reason for these small numerical differences.

It's not perfect, but the pipeline is at least checking that that the functionality of ignoring darks/flats runs without error. Let me know if the lack of sums/means being performed on the intermediate datasets are a concern, and if so, maybe something else can be done in the test to compensate?

PR #166, opened 2023-08-04
Add loader param for ignoring projections

### LINKED ISSUE #161: Functionality to skip specific projections
In Savu this functionality was implemented in the base class for reconstruction methods, so perhaps the analogous place in httomo would be in the recon wrapper class?

yousefmoazzam:
Fixes #161

Specifying a list of individual indices is supported, as well as a range of indices via start and stop values, similar to ignoring darks/flats.

TODO list:
- [x] Add parameter for ignoring projections
- [x] Add docs for new loader parameter
- [x] Add basic error checking that the provided projection indices to ignore in the YAML are indeed referring to projections
- [x] Add loader test to verify that the number of returned projections reflects the number that should have been skipped
- [x] OPTIONAL: Add test for `chunk.get_data_bounds()` to ensure data bounds get calculated correctly when running with multiple MPI processes (by mocking the return of `Comm.allgather()`)
- [ ] ~OPTIONAL: Implement for data that has been read as sinograms, and data that has been read along the third dimension~

### DIFF
--- a/docs/source/reference/loaders.rst
+++ b/docs/source/reference/loaders.rst
@@ -26,6 +26,10 @@ and to specify a range of indices to exclude, use the :code:`batch` subfield.
 These subfields can both be used at the same time to select batches of
 darks/flats/projections as well as individual darks/flats/projections to ignore.
 
+.. note:: The range provided to the :code:`batch` option via :code:`start` and
+          :code:`stop` is *inclusive*; ie, the :code:`stop` value is *included*
+          in the index values to ignore.
+
 It is also possible to ignore *all* darks and/or flats by setting the value of
 either :code:`ignore_darks` or :code:`ignore_flats` to :code:`true`. However,
 ignoring all *projections* is not yet supported.
--- a/docs/source/reference/loaders.rst
+++ b/docs/source/reference/loaders.rst
@@ -3,31 +3,55 @@
 HTTomo Loaders
 --------------
 
-Ignoring darks and/or flats
-===========================
+Ignoring darks, flats, or projections
+=====================================
 
-There are situations where certain dark or flat fields that have been collected
-in the data need to be excluded from processing. The standard tomography loader
-in HTTomo offers the functionality to exclude darks and/or flat field images
-based on their index in the dataset they are contained in. The images can be
-excluded by specifying either:
+There are situations where certain dark field, flat field, or projection images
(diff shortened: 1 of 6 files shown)


yousefmoazzam:
For anyone wondering why changes to `chunk.save_data_parallel()` were needed in 6f21c57, hopefully the following will explain the intention.

Consider 180 projections being split between 2 processes, so:
- rank 0 has 90 projections, indices 0 to 89
- rank 1 has 90 projections, indices 90 to 179

Then, 5 projections are ignored (say 85 to 89), all of which belong to rank 0, whereas rank 1 keeps all 90 of its projections.

Now the situation is
- rank 0 has 85 projections, indices 0 to 84
  - indices 85 to 89 have been removed
- rank 1 has 90 projections, indices 90 to 179

The assumption that the full data length in the slicing dimension (now 85 + 90 = 175) can be divided by the number of processes, rounded, and then used as a "step" to calculate the bounds of the data that each process has is now incorrect. Ie, the original code for calculating the data bounds when saving: https://github.com/DiamondLightSource/httomo/blob/f40f2744698241790ec3a0c2ffe9e55f704a6f30/httomo/data/hdf/_utils/chunk.py#L76-L80

would perform the following calculations to figure out the bounds of the data
that each process has:
```python
>>> nproc = 2
>>> length = 175
>>> round((length / nproc) * 0) # rank 0 start
0
>>> round((length / nproc) * (0 + 1)) # rank 0 stop
88
>>> round((length / nproc) * 1) # rank 1 start
88
>>> round((length / nproc) * (1 + 1)) # rank 1 stop
175
```

which leads to:
- rank 0 thinking its bounds to _write_ to the array in the hdf5 dataset are from indices 0 to 87
- rank 1 thinking its bounds to _write_ to the array in the hdf5 dataset are from indices 88 to 174

which is incorrect.

Instead, now that projections from indices 85 to 89 are removed:
- rank 0 should write its 85 projections (original indices 0 to 84) to bounds 0 to 84 in the hdf5 dataset
- rank 1 should write its 90 projections (original indices 90 to 179) to bounds 85 to 174 in the hdf5 dataset

yousefmoazzam:
Changes here are outdated, this feature will need to be added to the new loader. Ie, issue #161 is outdated, and issue #221 needs to be addressed with a separate PR.

Issue #178, opened 2023-09-14
Refactoring decorators


dkazanc:
The decorators being in httomolibgpu library pose few problems:
- multidata I/O needs to be dealt with in httomo (temporary solution)
- Adding methods directly requires either creating decorators  for a new module or creating a library file. Creating wrappers in httomolibgpu to add a new library goes against the idea of adding backends **directly** 

It has been decided to:

-  place all information about the backends in the local to httomo library file (similar to [tomopy](https://github.com/DiamondLightSource/httomo/blob/main/httomo/methods_database/packages/external/tomopy/1.14/tomopy.yaml) library) and remove decorators from httomolibgpu 
- initiate a new syntax that would direct a **single** memory estimation function, for instance
```
paganin: 
  memory: 
    arrays: [tomo, tomo]
    multipliers: [2, 1]
    methods: [none, Fourier_real]
```
or
```
normalize: 
  memory: 
    arrays: [tomo, flats, darks]
    multipliers: [2, 1.1, 1.1]
```
- Move MemoryHook tests into Httomo

PR #181, opened 2023-09-19
Wrappers unification and decorators refactoring [WIP]

dkazanc:
#178
* one BackendWrapper that would work for all backends
* retiring decorators from httomolibgpu
* new library files structure and unified functions to calculate max slices


dkazanc:
Hi @yousefmoazzam, before I moved too far away I thought it would be good to let you know where I'm heading with this.

1. I unified all wrappers in one single simple [BackendWrapper](https://github.com/DiamondLightSource/httomo/blob/0fb5fbbd210b3eb188a9e1df7392e19ede1c47e2/httomo/wrappers_class.py#L279) that deals with all the cases. Hopefully it makes easier adding new packages as it should cover them as well. 
2. I have a modified [library file](https://github.com/DiamondLightSource/httomo/blob/library/httomo/methods_database/packages/external/httomolibgpu/1.2/httomolibgpu.yaml) for httomolibgpu, please note the `method` section. 
3. In httomo we have a [function](https://github.com/DiamondLightSource/httomo/blob/0fb5fbbd210b3eb188a9e1df7392e19ede1c47e2/httomo/task_runner.py#L791) that calculates max slices which relies on `method` and `multipliers`. Specifically if the method is chosen as `direct`, then we do [simple arithmetic](https://github.com/DiamondLightSource/httomo/blob/0fb5fbbd210b3eb188a9e1df7392e19ede1c47e2/httomo/task_runner.py#L830) "in house" by using multipliers. If the method is `module` then we need to use a function from the module that uses more complicated arithmetic, similar to `_calc_max_slices` functions in httomolibgpu.
Here I'm not so sure where that module for calculating the memory should be. May be it should be a part of httomo actually, which makes easier for adding new packages which require more complicated memory estimation. So we do things 100% "in-house". 
4. I also changed the idea behind MemoryHook test which are now part of **test_backends**, see [normalisation](https://github.com/DiamondLightSource/httomo/blob/0fb5fbbd210b3eb188a9e1df7392e19ede1c47e2/tests/test_backends/test_httomolibgpu.py#L98) tests. We now comparing the  memory usage (hooked and estimated) and not slices, which is more comprehensive to me at least. 

Sorry, lots of changes to grasp. Please let me know if you have any questions. Thanks

dkazanc:
@yousefmoazzam just to add point 4. I think it is worth having modules to estimate memory local to httomo since:
1. this need is local to httomo and not the backend package
2. makes easier integrate any other package. 

So I'm currently testing the idea of paganin filter estimator being here:
`/httomo/httomo/methods_database/packages/external/httomolibgpu/memory_estimators/prep/phase.py`

dkazanc:
added new attribute `output_dims_change` which helps to work around changing data dimensions for some modules. So far implemented without connection with algorithm parameters (although it can be, e.g. padding, recon_size, etc.). Works uniformly across all backends so TomoPy "hacky" output shape recon is well generalised with it. 
```
recon:
  algorithm:
    FBP:
      pattern: sinogram
      output_dims_change: True
      implementation: gpu_cupy
      memory_gpu:
        - datasets: [tomo]
        - multipliers: [None]
        - methods: [module]
```

PR #182, opened 2023-09-28
Add pre-processing stage to pipeline

yousefmoazzam:
Main changes:
- add loader and pre-processing stages to pipeline YAML files via multiple "documents" within the YAML file
- enable dezinging (optional) and centering (mandatory) as part of the pre-processing stage
- try out defining a class for reading pipeline YAML files, as a test candidate for some refactoring
- refactor `test_yaml_checker.py` such that each individual check can be tested in isolation

### DIFF
--- a/httomo/cli.py
+++ b/httomo/cli.py
@@ -11,6 +11,7 @@ import httomo.globals
 from httomo.logger import setup_logger
 from httomo.task_runner import run_tasks
 from httomo.yaml_checker import validate_yaml_config
+from httomo.yaml_loader import YamlLoader
 
 from . import __version__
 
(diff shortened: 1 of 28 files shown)


yousefmoazzam:
@dkazanc  Note that the 360 pipeline is still not working with these changes so far. The `find_center_360` method works fine in the pre-processing stage (with some modifications to the config in `samples/pipeline_template_examples/DLS/02_i12_360scan_pipeline.yaml`), but the issue is then that the memory estimation for the stitching method `sino_360_to_180` needs to take into account the `overlap` value produced by `find_center_360`.

GPU memory estimation has had changes in httomo in the `gpuloop` branch which are not in this branch, so it probably makes sense to quickly review if people are ok with what is in this branch before making it work with the new way of doing memory estimation. (Namely, pipeline YAML file structure has changed a bit to allow for the pre-processing stage, and the `PipelineReader` class has been introduced).

Issue #103, opened 2023-04-05
TODO for the release


dkazanc:
In preparation for our first httomo/httomolib release, the following things needs of doing in order of priority:

- [x] Documentation for httomo that includes TomoPy yaml templates and links to TomoPy functions. We should add info how to run httomo and start documenting its functionality. 
- [x] [Yaml checker](https://github.com/DiamondLightSource/httomo/pull/101). This could be quite extensive and active development which will evolve with the feedback from the users. We will roll out the basic version of that to start with. 
- [x] Launcher. As the GPU development still in progress. The simple CPU launcher will suffice. 
- [x] [Logger](https://github.com/DiamondLightSource/httomo/pull/95)
- [x] Documentation for httomolib
- [x] Documentation for httomo that includes httomolib yaml templates and links to httomolib functions.
- [ ] Tests for all (preferably) TomoPy [templates](https://github.com/DiamondLightSource/httomo/tree/main/templates/tomopy). 

**Other pending but more "after release" issues**
- [ ] Inspection of padding (does it work?) and integration it into the meta as a decorator, similar to `pattern`
- [x] [Paganin filter reimplementation](https://github.com/DiamondLightSource/httomolib/issues/90).
- [ ]  Data stitching in httomolib
- [ ] Helical recon
- [ ] 4D data
- [ ] ....

PR #183, opened 2023-10-03
enabling padding

dkazanc:
#103 
Automation of padding is required for 3D methods to work smoothly in httomo 

- [x] Auto calculation of padding based on the parameters of the method
- [x] Enabling padding to be an attribute for every section based on which the data in blocks AND chunks will be padded
- [x] Enable padding to work for the serial run
- [ ] Enable padding to work for the parallel run
- [ ] Adding tests for padding


dkazanc:
OK, so serial padding seems to work as expected. I worked with the median filter, but in principle it can be enabled with more memory hungry methods like FBP in order to check the slicing is performed correctly for every block. It is nice to see that the boundary effects on blocks disappear as the padding applied (look for the horizontal line). 
no_padding
padded

dkazanc:
the serial run doesn't work, but not because of how padding implemented for blocks, It is how unpadding is performed (or not) for chunks. It doesn't seem to crop the padded chunks at all after padding them by looking at the dimensions. This all needs to be checked.

dkazanc:
closing this PR as the development moved to other branches.

PR #185, opened 2023-10-20
Add devcontainer

yousefmoazzam:
The idea is to provide a development environment for httomo that is easily reproducible, for both testing and development purposes.

Containers are one such tool that provide such reproducibility, and modern IDEs such as [VSCode](https://code.visualstudio.com/docs/devcontainers/containers) and [PyCharm](https://www.jetbrains.com/help/pycharm/connect-to-devcontainer.html#create_dev_container_inside_ide) support the ability to develop inside a container, while still being able to use all the features that the IDE provides when developing normally/outside of a container (such as autocompletion, debugging, Ctrl-click to jump to a definition, etc).

- [x] add Dockerfile for base container image with all deps of httomo, httomolib, and httomolibgpu installed
- [x] add basic config for dev container image
- [x] add support for VSCode
- [ ] add support for PyCharm
- [ ] add documentation on how to use the dev container


yousefmoazzam:
Adding support for PyCharm has been put on hold for the moment, as the [Dev Container](https://plugins.jetbrains.com/plugin/21962-dev-containers) plugin by JetBrains is unfortunately only available for the professional edition of PyCharm (ie, the paid edition of PyCharm).

There does appear to be some information [in the PyCharm docs](https://www.jetbrains.com/help/pycharm/connect-to-devcontainer.html#start_from_gateway) that using dev containers could be possible with the [JetBrains Gateway](https://www.jetbrains.com/remote-development/gateway/) program, but it's hard to tell if this only works with certain JetBrains IDEs.

PR #186, opened 2023-10-26
Improve estimation of FFT plans in GPU methods

yousefmoazzam:
Methods using FFTs to add improved FFT plan estimation to:
- [x] FBP
- [ ] remove_stripe_ti
- [x] paganin_filter_tomopy
- [ ] paganin_filter_savu

### DIFF
--- a/httomo/cufft.py
+++ b/httomo/cufft.py
@@ -1,0 +1,128 @@
+import ctypes
+import sys
+from enum import Enum
+
+
+if 'linux' not in sys.platform:
+    raise RuntimeError('Linux is currently the only supported platform')
+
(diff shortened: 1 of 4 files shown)


yousefmoazzam:
Note that FBP estimation is still under what CuPy reports by a small margin, so a bit more digging needs to be done into the filtering in that method.

dkazanc:
@yousefmoazzam can we close this as not planned as it seems that `httomo/cufft.py`  been taken for correcting FBP estimation elsewhere?

yousefmoazzam:
Yeah that's fine, we've taken `cufft.py` from it in #360 and that's all we need from this one

PR #187, opened 2023-11-01
 Separate slice indices and block cropping indices into functions 

jessicavers:


jessicavers:
I am merging these changes in case they are useful. So that people are aware, the branch had 21 failing tests of 143 prior to these changes. These changes don't appear to break any further tests

PR #190, opened 2023-12-05
Refactor task runner

jessicavers:


dkazanc:
I think it is ready to be merged to `gpuloop`. I've added more tests for DIAD and 360 pipelines recently. All seem to work as expected. Tests (404 of them!) pass.

httomo/data/hdf/_utils/reslice.py
```diff
@@ -181,4 +192,7 @@ def single_sino_reslice(
     )
 
     if mpiutil.rank == 0:
+        assert recvbuf is not None
         return recvbuf.reshape((data_shape[0], data_shape[2]))
+    else:
+        return None
```
yousefmoazzam, line 198: Naively, the additions here don't look like they change the behaviour of the function:
- I think python functions implicitly `return None` if they reach the end of a function and haven't hit a `return` statement
- asserting that `recvbuf is not None` for rank 0 is sort of made redundant by the if block at line 174 which assigns `recvbuf` to something

If I get rid of those additions and run the `test_does_reslice_when_needed` test locally it passes for me, could you run locally on your end to confirm if those changes are needed or not?

If not, I suppose it's still fine to leave in, so feel free to take it or leave it.

```suggestion
    if mpiutil.rank == 0:
        return recvbuf.reshape((data_shape[0], data_shape[2]))
```

tests/runner/test_task_runner.py
```diff
@@ -0,0 +1,377 @@
...
+    if filebased:
+        reslice_filebased.assert_called_once_with(
+                                                  ANY,
+                                                  1, 
+                                                  2,
+                                                  ANY,
+                                                  10,
+                                                  10,
+                                                  t.comm,
+                                                  reslice_dir)
```
yousefmoazzam, line 276: I think `ANY` was used in the assertion for the `data` parameter passed to `reslice_filebased()` rather than `dummy_dataset.data` because `dummy_dataset.data` actually gets set to the mock object during execution of the test due to:
- the patching of `reslice_filebased()`
- and the fact that in the task runner, the `data` attribute on the `DataSet` object is assigned to the return value of `reslice_filebased()` https://github.com/DiamondLightSource/httomo/pull/190/files#diff-d0196284273478c424c5d41d96d3d67518f3a6e9cda84e282da786d380407fe9R122
- hence, trying to assert that `dummy_dataset.data` was passed to `reslice_filebased()` doesn't work, because it's not a numpy array, but the mocked object instead

But the value for the `angles` parameter being passed to `reslice_filebased()` can still be asserted to be `dummy_dataset.angles` without issue since that doesn't get reset at any point during the execution of the test.

It makes sense to assert what we can, so asserting the angles should be done if it's simple enough, could you check if the `test_does_reslice_when_needed` test passes when using `dummy_dataset.angles` instead of `ANY`?

Also, some weird whitespace stuff going on here :stuck_out_tongue:

```suggestion
        reslice_filebased.assert_called_once_with(
            ANY,
            1, 
            2,
            dummy_dataset.angles,
            10,
            10,
            t.comm,
            reslice_dir
        )
```

httomo/data/hdf/_utils/save.py
```diff
@@ -4,14 +4,15 @@
...
 from httomo.utils import Colour, log_once
 
 
 def intermediate_dataset(
     data: numpy.ndarray,
     run_out_dir: Path,
-    loader_info: LoaderData,
+    angles: numpy.ndarray,
+    detector_x: int,
+    detector_y: int,
```
yousefmoazzam, line 15: I think the docstring needs to be updated accordingly

httomo/ui_layer.py
```diff
@@ -0,0 +1,112 @@
...
+                    if isinstance(value, str) and value is not None:
+                        if value.find('${{') != -1:                            
+                            result_extr = re.search(r"\{([A-Za-z0-9_.]+)\}", value)
+                            internal_expression = result_extr.group(1)
+                            (ref_id, side_str, ref_arg) = internal_expression.split(".")
+                            # lets find the referred id in "side_outputs_collect"
+                            for items in side_outputs_collect:
+                                if items[1] == ref_id:
+                                    # refer to methods_list[items[0]-1]
+                                    task_conf['parameters'][key] = OutputRef(methods_list[items[0]-1], ref_arg)
```
yousefmoazzam, line 85: Maybe not now, but at some point, it may be worth refactoring this to make it easier to read/understand.

Since this looks related to checking the validity of references to side outputs, extracting the logic to a separate function during refactoring may also prove useful for the discussion point we had the other day about wanting the YAML checker to check the validity of references in a pipeline YAML file?

tests/test_pipeline.py
```diff
@@ -337,187 +616,83 @@ def test_diad_testing_pipeline_output(
...
+#     log_contents = _get_log_contents(log_files[0])
 
-    assert (
-        "ERROR | Please provide `start`, `stop`, `step` values"
-        " when specifying a range to peform a parameter sweep over."
-    ) in log_contents
+#     assert (
+#         "ERROR | Please provide `start`, `stop`, `step` values"
+#         " when specifying a range to peform a parameter sweep over."
+#     ) in log_contents
```
yousefmoazzam, line 698: If we're not interested in thinking about sweep functionality right now, it may be better to just remove this entirely rather than commenting out (other commented-out sweep tests like `test_sweep_pipeline_with_save_all_using_mpi` have been removed in this PR anyway)

yousefmoazzam:
A general note: this PR contains a lot of changes from several different contributors, and the variations in the code formatting becomes apparent quite early on when reading through the changes. As silly as it sounds, after a few minutes of reading, it can start to get difficult to focus on what the code is doing when the formatting changes significantly.

I think code reviews are very useful to do, and so making them as easy as can be possible would be great!

Additionally, having a code style in place can also help external people contribute, because they can just follow the code style rules and not need to worry about what is good/bad. If a commonly used code formatting tool is chosen then it also means that the code will be formatted in a way that other people would be used to, and again, reducing the barrier-to-entry for contributing.

With these things in mind, I think it might be good for us to pick a concrete point in the future to start using a code formatter, like we would do for any other goal/milestone in the development of software. I don't mind which formatter, nor really the specifics of each rule within the formatter, as long as it's used consistently.

The point in time time could be anything:
- after `refactor-task-runner` is merged into `gpuloop`
- after `feature-transparent-file-store` is merged into `gpuloop`
- after everything is merged into `main` and before the initial release
- after everything is merged into `main`, and after the initial release
- whatever else would be suitable

but I think picking a time would be useful (if we do decide to go with a formatter), so then the idea doesn't get left behind.

It's just a thought, it'd be interesting to hear what others think :slightly_smiling_face:

httomo/ui_layer.py
jessicavers: I agree with this. I was looking at this same part yesterday, and thinking of refactoring to see if that could be tested in a more isolated way.

jessicavers:
I think that using a consistent formatter has been mentioned before. I am in general agreement and think it would be good to decide on a time to add also.

httomo/data/hdf/_utils/reslice.py
dkazanc: Although mypy is now marks the whole function red as it explicitly missing the return statement...

yousefmoazzam: This is fair enough, I use `pyright` instead of `mypy` for type-checking (because I already use `pyright` as a language server anyway) and `pyright` doesn't complain to me about a missing return statement, they likely have different rules :shrug:

httomo/ui_layer.py
dkazanc: yes, good point. Feel free to move to a separate function that assigns the side_outputs references. @jessicavers I guess you could do that in your branch for yaml checker?

PR #191, opened 2023-12-07
WIP for transparent file store

team-gpu:
This PR is work in progress to have a file-backed store for handling datasets that are larger than the CPU memory, reading/writing blocks to/from it as needed. It only goes to disk when there is not enough memory - otherwise it stays in-memory (for performance reasons). 

The outline is largely there - but there are failing tests with MPI and reslicing that still need to be fixed.


yousefmoazzam:
I believe that this PR is stale and has been replaced by #198, feel free to comment/reopen if I'm mistaken.

PR #196, opened 2023-12-18
Enabling new TaskRunner with cli changes

dkazanc:
- Enable new TaskRunner
- Fix tests
- Remove unused files


dkazanc:
- The runner has been re-enabled to work with the new YAML pipelines and also pipelines written in Python. I've added tests for all of them and they pass. 
- save_result as a module argument to any template has been added and tests fixed to accommodate it. 
- the yaml merger has been re-written as the previous version was creating a separator, resulting in an error
- various minor fixes

The only test that fails for me in this PR is: **test_does_reslice_when_needed** , @yousefmoazzam were you looking into that one?

PR #200, opened 2024-01-16
use filestore in all processes if any of them runs out of memory

team-gpu:
This fixes the issue that if only some processes run out of memory, the file-based store should be used by *all* processes.


tests/data/test_dataset_store.py
```diff
@@ -364,107 +366,97 @@ def test_reslice_single_block_single_process(
...
     if comm.rank == 0:
-        np.testing.assert_array_equal(
-            block.data,
-            global_data[:, 1:3, :],
-        )
-    else:
+        assert writer.chunk_index == (0, 0, 0)
+        assert reader.chunk_index == (0, 0, 0)
+        np.testing.assert_array_equal(block.data, global_data[:, 1:3, :])
+
```
yousefmoazzam, line 453: Tiny detail, but an unnecessary newline is here between the if/elif I think?

```suggestion

```

PR #201, opened 2024-01-16
Yaml checker changes

jessicavers:


httomo/yaml_checker.py
```diff
@@ -325,3 +348,16 @@ def validate_yaml_config(
     )
     _print_with_colour(end_str, colour=Colour.BVIOLET)
     return True
+
+
+class UniqueKeyLoader(yaml.SafeLoader):
```
jessicavers, line 369: For clarity and background, this class is building on the yaml.SafeLoader and is used from a post at https://stackoverflow.com/questions/33490870/parsing-yaml-in-python-detect-duplicated-keys

dkazanc:
- just few mypy complains about path/str types 
- unified the yaml loader so that one function used everywhere (there was a duplication)
- and most importantly few tests were failing for me as the version of httomo was extracted from by "dirty" editable (dev) version by this [function](https://github.com/jessicavers/HTTomo/blob/91c89e324406e280bbf3b75672facc9fd338d4b8/httomo/yaml_checker.py#L260) . I decided to unify the idea of getting the version the same way for not just the external files but for httomo as well.

dkazanc:
@jessicavers shall we enable the checker in `cli` before merging this may be? 
At some point we also need to update the docs about the yaml checker. Don't know if it should be mentioned as a separate entity as it becomes a part of the runner. May be no need.

jessicavers:
> * just few mypy complains about path/str types
> 
> * unified the yaml loader so that one function used everywhere (there was a duplication)
> 
>* and most importantly few tests were failing for me as the version of httomo was extracted from by "dirty" editable (dev) version by this [function](https://github.com/jessicavers/HTTomo/blob/91c89e324406e280bbf3b75672facc9fd338d4b8/httomo/yaml_checker.py#L260) . I decided to unify the idea of getting the version the same way for not just the external files but for httomo as well.


Thank you for the mypy updates.
That seems useful to unify the version in the same way as for the other files.

I meant to ask you about the yaml loader. Yes, there was duplication. I do have a question/two about that. Not vital, more general arrangements.

There was a yaml loader used in the yaml checker test multiple times ([eg](https://github.com/DiamondLightSource/httomo/blob/91c89e324406e280bbf3b75672facc9fd338d4b8/tests/test_yaml_checker.py#L42)). I created a fixture in [conftest.py](https://github.com/DiamondLightSource/httomo/blob/91c89e324406e280bbf3b75672facc9fd338d4b8/tests/conftest.py#L321) to load the yaml. Is it recommended I import the ui_loader [_yaml_loader](https://github.com/DiamondLightSource/httomo/blob/91c89e324406e280bbf3b75672facc9fd338d4b8/httomo/ui_layer.py#L109) to use in these tests instead? Or it's better to have the loader for a separate test to be separated, in case of error with the loader in main code? I wasn't certain.

I also had a question about what we return. [_yaml_loader](https://github.com/DiamondLightSource/httomo/blob/91c89e324406e280bbf3b75672facc9fd338d4b8/httomo/ui_layer.py#L109) returns a list, meaning after it is used at the moment, we'd select the [first item](https://github.com/DiamondLightSource/httomo/blob/91c89e324406e280bbf3b75672facc9fd338d4b8/httomo/ui_layer.py#L39) in the list. In the other function I had returned the first item, to avoid that, and also to specify the returned type. That wasn't exactly the best way that I arranged it, I think, as it included that one list step, with a slightly different purpose, into the loading function. I don't think my arrangement is what we should use, I just wanted to highlight this, I guess, to hear your view.

jessicavers:
> @jessicavers shall we enable the checker in `cli` before merging this may be? At some point we also need to update the docs about the yaml checker. Don't know if it should be mentioned as a separate entity as it becomes a part of the runner. May be no need.

Yes, that seems good to wait for that until merging. 
Good point, yes, I'll need to update the documentation.

dkazanc:
thanks, yaml loader in [conftest.py](https://github.com/DiamondLightSource/httomo/blob/91c89e324406e280bbf3b75672facc9fd338d4b8/tests/conftest.py#L321) can probably stay as it is, no need to rely on the one in ui_layer. What the loader returns can be changed I guess, but it needs to be done for ui_layer and the checker. I implemented it to return a list just because I was checking the list contents. 

Documentation can be probably postponed for dedicated work on it. The checker cam be added to cli and we're OK here I guess.

httomo/yaml_checker.py
```diff
@@ -216,114 +130,165 @@ def check_methods_exist_in_templates(conf: PipelineConfig) -> bool:
...
-            method_dict = yaml.load(template, Loader=yaml.FullLoader)[0]
-            template_yaml_data_list.append(
-                next(iter(method_dict.values()))
+    template_yaml_conf = _get_template_yaml_conf(conf)
+    for method_dict in conf:
+        yml_method_list = [
+            yml_method_dict
+            for yml_method_dict in template_yaml_conf
+            if (yml_method_dict["method"] == method_dict["method"])
+            & (yml_method_dict["module_path"] == method_dict["module_path"])
```
yousefmoazzam, line 157: A nitpicky point, but I think in python, the `&` operator is a [bitwise operator](https://docs.python.org/3/reference/expressions.html#binary-bitwise-operations) and is intended to be used on bit values.

Possibly due to some type coercion happening or something it still works as intended, but it may be worth changing the `&` to `and` to signal the intent that the code isn't expecting to be working on bit values here (at least, it doesn't look like it's expecting bit values, but please do correct me if I'm wrong)?

```suggestion
            and (yml_method_dict["module_path"] == method_dict["module_path"])
```

jessicavers: Thank you for checking this and making the suggestion. Completed in [71641dd](https://github.com/DiamondLightSource/httomo/pull/201/commits/71641dd9f80f66081ba0c954586eaec17467ee42)

jessicavers:
Okay. Following that previous comment, I have updated the _yaml_loader here [679ca6ab](https://github.com/DiamondLightSource/httomo/commit/679ca6abdbe29f71218b3a14f37833474c4c7538), to return the PipelineConfig. The ui_layer and yaml_checker have both been updated as mentioned. I also decided to rename it to yaml_loader, removing the starting underscore, as I am importing and using this function elsewhere. I hope that is suitable. 

The _yaml_check is incorporated, but I have not had time to incorporate the logging file capability we had discussed, so the error messages are more hidden from the user. When added, that should improve the experience a lot.

httomo/cli.py
```diff
@@ -59,11 +60,11 @@ def run(
...
 ):
     """Run a pipeline defined in YAML on input data."""
-    
-    # First we need to validate yaml configuration file if there are any errors
-    # TODO: with new yaml syntax check yaml is not fully working.
-    # Need to re-enable that: 
-    # _check_yaml(yaml_config, in_data_file)
+    if yaml_config.suffix == ".yaml":
+        try:
+            assert _check_yaml(yaml_config, in_data_file) == True
```
yousefmoazzam, line 65: A nitpick, feel free to take it or leave it. I think the checking of equality of the returned boolean value of `_check_yaml()` is unnecessary here? 

Ie, I think this is enough:
```python
assert _check_yaml(yaml_config, in_data_file)
```

httomo/yaml_checker.py
```diff
@@ -216,114 +130,208 @@ def check_methods_exist_in_templates(conf: PipelineConfig) -> bool:
...
-def _get_yaml_templates(modules: List, methods: List, packages: List) -> List:
+def _get_template_yaml(conf: PipelineConfig, packages: List) -> List:
     """
-    Helper function that fetches YAML template files associated with methods
+    Helper function that fetches template YAML file names associated with methods
     passed.
     """
     parent_dir = os.path.dirname(os.path.abspath("__file__"))
     templates_dir = os.path.join(parent_dir, "templates")
     assert os.path.exists(templates_dir)
```
yousefmoazzam, line 326: Thanks for replacing the various `assert` statements in this file with printing of a helpful error message to describe the issue before returning a `bool` :slightly_smiling_face: It makes it easier for anything calling the top-level `validate_yaml_config()` function to simply:
- execute that function
- let any failures be printed
- and finally check the `bool` value that was returned

There was something that came to mind when I thought about the launcher python script calling `validate_yaml_config()` and any error handling it would have to do to terminate the program before the job was submitted.

This `assert` looks like it might be the only place now where the `validate_yaml_config()` function can possibly fail and a nice error message _isn't_ printed.

Ie, even if the caller does something like:
```python
try:
    validate_yaml_config(my_yaml)
except AssertionErrror as e:
   print(f"YAML check failed with unexpected error: {e}")
```

and the assertion failed, the `e` would have no error message within to use for printing, because no error message is reported when the error is raised.

Do you think it would make sense to given an error message in case the assertion fails, so then a caller can use a try/except as above to then handle the error and avoid displaying the stacktrace to the user?

Ie, something like:
```python
assert os.path.exists(templates_dir), "Dev error: expected YAML templates dir to exist"
```

which could then be nicely handled by the caller in the unlikely event that the templates directory doesn't exist for whatever reason?

httomo/cli.py
jessicavers: Thank you for highlighting this, updated as suggested in [822a202](https://github.com/DiamondLightSource/httomo/pull/201/commits/822a2023491ab8d0b01d76577c15590e6a79b0ae). However, then this piece of code was removed following discussion on at which point yaml checker will be run.

httomo/yaml_checker.py
jessicavers: Thank you for reviewing. For the error messages, I had been following the format which Naman had set up. This assert statement does not take into account logging and error messages. This was a short term solution under time constraint to halt the later run if there was an error inside the yaml file. It definitely is sensible to include clearer logging capability, as had previously discussed, but this hasn't been included in this branch.

yousefmoazzam: > It definitely is sensible to include clearer logging capability, as had previously discussed, but this hasn't been included in this branch.

To be clear, I'm not suggesting adding logging capabilities or anything, I'm suggesting to add a message to the `assert` that it should return if it fails, a one-line change like:
```suggestion
    assert os.path.exists(templates_dir), "Dev error: expected YAML templates dir to exist"
```

so then callers of the function can report what went wrong without displaying a nasty stacktrace to the user.

yousefmoazzam:
Fixes #281

httomo/yaml_checker.py
jessicavers: Thanks for clarifying, added in [f53709a](https://github.com/DiamondLightSource/httomo/pull/201/commits/f53709ac6350898dcfe874961241d98e3a722f84)

yousefmoazzam (commented):
All tests pass for me locally, and using the `check` command seems to work fine.

I found one bug, see #302, but I'm fine with that fix being done later if it's better to get this merged sooner rather than later.

PR #202, opened 2024-01-19
Refactor `StandardTomoLoader` and associated tests

yousefmoazzam:
An assortment of small improvements:
- reduce duplicate boilerplate code in loader tests for creating loader object
- modify relative paths to work regardless of current working directory when executing loader tests
- store a file handle to the hdf5 file on the loader so then the file doesn't need to be opened more than once
- implement the `finalize()` method on `StandardTomoLoader` to make it satisfy to the `DataSetSource` interface once more

### DIFF
--- a/httomo/runner/loader.py
+++ b/httomo/runner/loader.py
@@ -230,7 +230,7 @@ class StandardTomoLoader(DataSetSource):
         return self._h5file[self._angles.data_path][...]
 
     def finalize(self):
-        pass
+        self._h5file.close()
 
 
 def get_darks_flats(
--- a/httomo/runner/loader.py
+++ b/httomo/runner/loader.py
@@ -103,6 +103,7 @@ class StandardTomoLoader(DataSetSource):
         self._angles = angles
         self._slicing_dim = slicing_dim
         self._comm = comm
+        self._h5file = h5py.File(in_file, "r", driver="mpio", comm = comm)
 
         self._data_indices = self._get_data_indices()
         self._global_shape = self._get_global_data_shape()
@@ -125,7 +126,7 @@ class StandardTomoLoader(DataSetSource):
         angles_arr = self._get_angles()
         darks_arr, flats_arr = get_darks_flats(darks, flats, comm)
 
-        dataset: h5py.Dataset = self._get_h5py_dataset()
+        dataset: h5py.Dataset = self._h5file[data_path]
         self._data = FullFileDataSet(
             data=dataset,
             angles=angles_arr,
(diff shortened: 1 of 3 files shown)


httomo/runner/loader.py
```diff
@@ -166,12 +167,16 @@ def __init__(
...
+        if slicing_dim != 0:
+            raise NotImplementedError("Only slicing dim 0 is currently supported")
+
         self._in_file = in_file
         self._data_path = data_path
         self._image_key_path = image_key_path
         self._angles = angles
         self._slicing_dim = slicing_dim
         self._comm = comm
+        self._h5file = h5py.File(in_file, "r", driver="mpio", comm = comm)
```
team-gpu, line 179: I assume the change to avoid the `mpio` driver for reading is done in the other branch?

httomo/runner/loader.py
```diff
@@ -166,12 +167,16 @@ def __init__(
         slicing_dim: Literal[0, 1, 2],
         comm: MPI.Comm,
     ) -> None:
+        if slicing_dim != 0:
+            raise NotImplementedError("Only slicing dim 0 is currently supported")
+
```
team-gpu, line 172: you should add a quick test for this - unless you fixed it in the previews branch to support all slicing dims?

httomo/runner/loader.py
```diff
@@ -261,27 +267,18 @@ def _calculate_chunk_shape(
...
     # NOTE: This method is largely copied from `load.get_data_indices()`; that function should
     # be removed in the future if/when `StandardTomoLoader` gets merged.
     def _get_data_indices(self) -> List[int]:
-        with h5py.File(self._in_file, "r", driver="mpio", comm=self._comm) as f:
-            if self._image_key_path is None:
-                data: h5py.Dataset = f[self._data_path]
-                return np.arange(data.shape[0]).tolist()
+        if self._image_key_path is None:
+            data: h5py.Dataset = self._h5file[self._data_path]
+            return np.arange(data.shape[0]).tolist()
```
team-gpu, line 279: These lines are not covered by tests (I ran a coverage). I think it's an important special case that you should cover.

team-gpu (changes requested):

httomo/runner/loader.py
yousefmoazzam: Done with [486c9e9](https://github.com/DiamondLightSource/httomo/pull/202/commits/486c9e99e795b629f92d360c0e56e13548093d6d)

httomo/runner/loader.py
yousefmoazzam: Good catch, thanks! I do remember noticing this when I was making changes on the `previewing` branch after I made this PR.

The work on the `previewing` branch included moving the data indices calculation out of `StandardTomoLoader` into a `Preview` object. At that time, I then added a test to the `Preview` object to handle the case when the image key path is `None`, see the `test_preview_with_no_image_key()` test in `test_loader.py` in the PR for adding previewing https://github.com/DiamondLightSource/httomo/pull/209/files#diff-93229d65c37eac8fb2c218f0a1f0ebeb3a2fadf848ffd6557bf1375e7942dac3R769. (Apologies, the link won't automatically jump to the highlighted function due to the file diff being too large to show by default, one needs to manually click to show the diff for the `test_loader.py` file for it to then jump to the highlighted test function).

So I'm going to leave out a test for the loader to handle that case in this PR, because the PR for previewing will tackle it, hope that makes sense.

team-gpu: sounds good

PR #205, opened 2024-01-22
Integration of CPU block-wise process in task runner

team-gpu:
This fully integrates block-wise CPU data processing into the `TaskRunner`, using the new dataset store and source/sink. In particular:

## Method Wrappers

* Renames `BackendWrapper` to `MethodWrapper` and moves them into their own directories (incl. their tests)
* Defines `MethodWrapper` as a pure interface (`Protocol`), and introduces `GenericMethodWrapper` for the base class implementing that interface. This is used for most methods.
* Specific child classes are needed for some method types, and the `@classmethod` `should_select_this_class` should be used to implement the condition when a particular wrapper should be used.
* The factory function `make_method_wrapper` is selecting the correct class based on this class method

### Stats Calculations

- modified to work with blocks and running averages
- MPI is used only at the end, on the last block in chunk
- MPI is factored out into the wrapper  - the method is plain and simple

### Images Wrapper

- modified to work with block-wise input
- added input parameter `offset` and using that in the wrapper (see [related PR](https://github.com/DiamondLightSource/httomolib/pull/6) which should be merged first, perhaps into a `library` branch while all this is in development)

### Save intermediate files wrapper
  
- implemented a blockwise method wrapper for this task
- also takes loader instance (for `detector_x` and `detector_y` data) + a reference to the previous method (to determine file name)

### Method Database

- added `default_save_results` option to DB, which is `True` for reconstruction methods but can be overwritten with the user parameter `save_results` -> please check default values in this PR

## Platform Sectioning

- renamed `PlatformSection` to `Section`, and only introduces a new section when there is a pattern change or when an output of a previous method is referenced in a method parameter
- when the loader can only provide a different pattern than the first method, it inserts an empty section, which will have the effect of causing a reslice directly after loading
- Uses `max_slices = 64` for CPU-only sections, with the option to configure this on the command-line

## Task Runner

- block split is now using a dataset source
- aggregator not needed anymore (uses a dataset sink)
- added a dummy sink for the last section which doesn't do anything

## Transformation Layer

- inserted pipeline transformation layer, running between UI layer and runner
- at the moment, inserts save intermediate files method wrappers as needed (depending on `save_results` flag or `--save-all`)
- removes file-saving work from from task runner completely - it simply executes methods

## UI Layer

- Revised and improved code, improved test coverage, and made output references more robust

## Tests

- All tests are passing, with almost all new code providing a 100% test coverage
- The FBP backend tests for memory estimation fail for some sizes - fixes for this are in another branch
- The end-to-end pipeline tests are mostly failing, due to the loader not yet supporting previews and not logging the same messages


httomo/runner/dataset.py
```diff
@@ -178,6 +179,11 @@ def is_full(self) -> bool:
         """Check if the dataset is the full global data"""
         return False
 
+    @property
+    def is_full(self) -> bool:
+        """Check if the dataset is the full global data"""
+        return False
+
```
yousefmoazzam, line 186: Potentially a duplicated method, the diff shows that `is_full()` is already there above the newly added lines?

httomo/runner/dataset.py
```diff
@@ -337,6 +346,11 @@ def is_full(self) -> bool:
...
```
yousefmoazzam, line 353: Another duplicated method?

httomo/runner/dataset_store_interfaces.py
```diff
@@ -48,68 +50,131 @@ class DataSetSource(Protocol):
...
+    """Interface for a DataSetSink that is store-based, i.e. where it's possible to construct
+       a reader from the same data store that this sink has been written first"""
+       
+    @abc.abstractmethod
+    def make_reader(
+        self, new_slicing_dim: Optional[Literal[0, 1, 2]] = None
+    ) -> DataSetSource:
+        """Method to make a source from this sink, which will read the data that was written,
+        possibly in a new slicing dimension"""
+        ...  # pragma: no cover
```
yousefmoazzam, line 146: Apologies for being dumb, I think we did discuss stuff in this area, but I'm still trying to understand things when reading through. I'm sort of struggling to understand the utility of the `StoreBasedDataSetSink` interface. 

`StoreBasedDataSetSink` adds a `make_reader()` method on top of `DataSetSink`, but `DataSetStoreWriter` already had a `make_reader()` method in its implementation of the `DataSetSink` interface. So I'm sturggling to see the difference between:
- `DataSetStoreReader` that already existed
- `StoreBasedDataSetSink` added in this change

as far as their usage in the task runner is concerned.

Also, given its name and docstring, `StoreBasedDataSetSink` seems to be somewhat tied to the data store, and the data store is a _particular implementation_ of the source and sink interfaces. In other words, `StoreBasedDataSetSink` is an interface (an abstraction), but is somewhat connected to or dependent on a specific implementation of other interfaces (a concretion), which seems a bit odd?

Any comments to help me reconcile my confusion would be great, and sorry again for being slow to grasp things!

httomo/runner/dataset.py
team-gpu: indeed - that was a merge error. Fixed.

httomo/runner/dataset.py
team-gpu: fixed

httomo/data/dataset_store.py
```diff
@@ -281,7 +280,7 @@ def _create_h5_data(
 
     def make_reader(
         self, new_slicing_dim: Optional[Literal[0, 1, 2]] = None
-    ) -> "DataSetStoreReader":
+    ) -> DataSetSource:
```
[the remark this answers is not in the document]
team-gpu: see my reply below...

httomo/runner/dataset_store_interfaces.py
team-gpu: The reason for this is in [task_runner.py](https://github.com/DiamondLightSource/httomo/pull/205/files/726c9ab12ec9eccab089cc59265a12b5673daa91#diff-d0196284273478c424c5d41d96d3d67518f3a6e9cda84e282da786d380407fe9R45) - it needs to call `make_reader` for processing the next section, and technically it doesn't need to know if that's a store based writer or whatever it is. All it needs to know is that it's a sink interface with a `make_reader` method.

However, note that this is still work in progress - you will see that  `TaskRunner` actually instantiates `DataSetStoreWriter` explicitly, so it knows about it anyway. So your point is valid if it stays that way - at the moment it's not actually hiding implementations for `TaskRunner`. But that might change before the PR is ready to merge - still thinking about how.

yousefmoazzam: Ok, this makes sense that, due to being a work-in-progress, there's still a concrete dependence of `DataSetStoreWriter` in `TaskRunner`.

I guess that an interface which adds a `make_reader()` method on top of `DataSetSink` would still be needed in the future when the task runner is able to work with _any_ writer as long as a reader can be created from the writer. 

Currently, the interface that outline this "writer + can create reader from" behaviour is called `StoreBasedDataSetSink`. Which is a fine name for now, because the one and only implementation of that interface (`DataSetStoreWriter`) is indeed a sink that's based on a store.

However, given that in the future, a sink implementing a writer interface that requires a `make_reader()` method does not need to be stored-based necessarily, I think it'd be worth changing the name `StoreBasedDataSetSink` to something that doesn't refer to "store".

team-gpu: Do you have a suggestion for how to name things?

yousefmoazzam: It's a very good question... :sweat_smile:

The fundamental difference between `DataSetSink` and `StoreBasedDataSetSink` is that the latter describes a sink that also allows the creation of a reader from it, isn't it? Maybe something like `ReadableSink`?

I'm not sure though, it's a name that makes some sense to me based on what behaviour the interface describes, but that doesn't mean it'd be sufficiently explanatory to others.

If we can't come up with a better name for it, I'm content with leaving it as `StoreBasedDataSetSink` for now, and to perhaps rename it in the future, possibly when we go over the data object hierarchy and rename stuff like `FullFileDataSet`.

team-gpu: done

PR #209, opened 2024-02-02
Add basic previewing functionality to `StandardTomoLoader`

yousefmoazzam:
Added:
- some httomo internal data structures for previewing: `Preview`, `PreviewConfig`, `PreviewDimConfig`
- functions that transform `preview` and `angles` parameter config into values of type `PreviewConfig` and `AnglesConfig`  respectively
- previewing functionality for projections + darks/flats in detector y and detector x dimensions (assuming unit-step)

Missing functionalities:
- previewing functionality for projections/darks/flats with **non-unit step** in all dimensions


httomo/runner/loader.py
```diff
@@ -151,6 +151,97 @@ class UserDefinedAngles(NamedTuple):
...
+                f"start={self.config.detector_y.start}, "
+                f"stop={self.config.detector_y.stop}"
+            )
+
+        if self.config.detector_x.stop > shape[2]:
+            raise ValueError(
+                f"Preview indices in det x dim exceed bounds of data: "
+                f"start={self.config.detector_x.start}, "
+                f"stop={self.config.detector_x.stop}"
+            )
```
team-gpu, line 197: can you add checks for `start >= stop` as well?

httomo/runner/loader.py
```diff
@@ -151,6 +151,97 @@ class UserDefinedAngles(NamedTuple):
...
+        return intersection.tolist()
+
+    def _calculate_global_shape(self) -> Tuple[int, int, int]:
+        if self._data_indices is None:
+            self._data_indices = self._calculate_data_indices()
+
+        return (
+            len(self._data_indices),
+            len(range(self.config.detector_y.start, self.config.detector_y.stop)),
+            len(range(self.config.detector_x.start, self.config.detector_x.stop)),
```
team-gpu, line 229:
```suggestion
            self.config.detector_y.stop - self.config.detector_y.start,
            self.config.detector_x.stop - self.config.detector_x.start,
```

httomo/runner/loader.py
```diff
@@ -197,11 +294,15 @@ def __init__(
...
 
-        dataset: h5py.Dataset = self._h5file[data_path]
+        dataset: h5py.Dataset = self._get_data()
         self._data = FullFileDataSet(
-            data=dataset,
+            data=dataset[
+                self._preview.config.angles.start : self._preview.config.angles.stop,
+                self._preview.config.detector_y.start : self._preview.config.detector_y.stop,
+                self._preview.config.detector_x.start : self._preview.config.detector_x.stop,
+            ],
```
team-gpu, line 305: I am quite certain this actually performs a full read operation from the file into RAM. The return type of dataset slicing operations `dataset[:, ...]` is a numpy array. If the data doesn't fit in memory, that's a problem. Ideally the data should sit in the file and only be read when a block is required. 

I understand that this probably needs to move offsets into `FullFileDataSet`....

team-gpu (changes requested):

httomo/runner/loader.py
yousefmoazzam: Thanks for noting this, the brain does unnecessarily complicated things sometimes! :laughing: Done in [0ebdedb](https://github.com/DiamondLightSource/httomo/pull/209/commits/0ebdedb11c9eac3b72e83ddadfe1d055b66caf11)

httomo/runner/loader.py
yousefmoazzam: Good idea, done in [ca9a0fa](https://github.com/DiamondLightSource/httomo/pull/209/commits/ca9a0fac3dddbfca0dc0578ad32d6158be54c713)

httomo/runner/loader.py
yousefmoazzam: I believe this particular instance of accidentally causing a read of the full data is resolved. But another round of reviewing will determine if the issue is gone for good, or if it's sprung up in a new place... :sweat_smile:

team-gpu: I wonder if we shouldn't put a test for that. I think it's possible - with an `assert isinstance(loader._data.data, h5py.Dataset)` or something like that. To make sure it's not an `ndarray` - this would also ensure that future refactorings don't accidentally create the same issue again.

yousefmoazzam: Thanks for the note, this has been done in [caa9d26](https://github.com/DiamondLightSource/httomo/pull/209/commits/caa9d260b2e1804b36cd58c6d2981e11dd0ef67e)

team-gpu (changes requested):

httomo/runner/loader.py
```diff
@@ -78,6 +78,102 @@ class UserDefinedAngles(NamedTuple):
...
+    ) -> None:
+        if config.stop > length:
+            raise ValueError(
+                f"Preview indices in {name} dim exceed bounds of data: "
+                f"start={config.start}, " f"stop={config.stop}"
+            )
+
+        if config.start >= config.stop:
+            raise ValueError(
+                f"Start index in preview indices for {name} dim is >= stop index: "
```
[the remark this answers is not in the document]
yousefmoazzam: Suggestions like this are really useful, when in the weeds of development it can be hard to come up with an error message that would make sense to users. I like your suggestion, I'll add that change!

httomo/runner/loader.py
```diff
@@ -78,6 +78,102 @@ class UserDefinedAngles(NamedTuple):
...
+                f"start={config.start}, stop={config.stop}"
+            )
+
+    def _calculate_data_indices(self) -> List[int]:
+        if self._image_key is not None:
+            indices = np.where(
+                self._image_key[:] == 0
+            )[0].tolist()
```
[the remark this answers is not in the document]
yousefmoazzam: See other comment about `numpy.nonzero()` returning a tuple of arrays, the same question applies here I guess.

team-gpu (approved):

httomo/loaders/standard_tomo_loader.py
```diff
@@ -0,0 +1,272 @@
...
+    data_path: str
+
+
+class UserDefinedAngles(NamedTuple):
+    start_angle: int
+    stop_angle: int
+    angles_total: int
+
+
+AnglesConfig: TypeAlias = Union[RawAngles, UserDefinedAngles]
```
team-gpu, line 27: I think it's better to define the types where the interface to the outside world is. That is, I'd move this with `make_loader` - or have the types in a separate file. Then this file feels more like one possible implementation of a loader - and a future different one would not have to include this file to get the types.

httomo/method_wrappers/rotation.py
```diff
@@ -68,7 +68,7 @@ def _build_kwargs(
         if not "ind" in dict_params or (
             dict_params["ind"] == "mid" or dict_params["ind"] is None
         ):
-            updated_params = {**dict_params, "ind": (dataset.global_shape[1] - 1) // 2}
+            updated_params = {**dict_params, "ind": (dataset.base.shape[1] - 1) // 2}
```
team-gpu, line 71: Why was this change needed again? Isn't the index the global shape's midpoint? And given that the rotation methods are project pattern - what's the difference between the old and the new?

team-gpu (changes requested):
Are the basic pipeline tests all passing now with these changes?

httomo/method_wrappers/rotation.py
yousefmoazzam: Ah, possibly this doesn't need to be done now. This change was done this commit [5b0a2b0](https://github.com/DiamondLightSource/httomo/pull/209/commits/5b0a2b04d349be75245dc872f8c1fb517c93d217), which came _before_ the merge I did of `feature/transparent-file-store` into `previewing` to make use of `FullFileDataSet.data_offset` [39567b0](https://github.com/DiamondLightSource/httomo/pull/209/commits/39567b0c8c7259a5592eb6a7730fa397a3365192).

This change was needed when I was still trying to use `FullFileDataSet.global_index` to do offsets for previewing, but quite possibly it's not needed now that we're using `FullFileDataSet.data_offset`.

I'll check if this is actually needed, thanks for the keen eye!

httomo/loaders/standard_tomo_loader.py
yousefmoazzam: Good point, I wasn't sure where these types should live, I think alongside `make_loader()` in `httomo/loaders/__init__.py` is a good idea.

httomo/method_wrappers/rotation.py
team-gpu: I get it - so I think `dataset.shape` should be enough here? 

Still, I think `FullFileDataSet.global_shape` should include the previewing / offsetting ...

yousefmoazzam: > Still, I think FullFileDataSet.global_shape should include the previewing / offsetting ...

The problem with this is that it causes this assertion to fail I think: https://github.com/DiamondLightSource/httomo/blob/dd193901bd4ae905a4df72cbbda5041af20c803c/httomo/runner/dataset.py#L44-L47

What happens is that:
- `data` is the h5py `Dataset` object
- `data.shape` is the shape of the full raw data
- whereas if we were to have `global_shape` be the previewed shape, this then is "smaller" than `data.shape` -> assertion fails

And it's a sensible assertion to be honest, so I didn't want to just remove it, but I guess we can also do that if that is an ok tradeoff to be able to have `FullFIleDataSet.global_shape` be the previewed shape?

I'm still not sure about `FullFileDataSet.global_shape` being the previewed shape, because in the context of a "full file", the global shape makes sense to be the "full shape" without any previewing, especially since `FullFileDataSet` doesn't know anything about previewing, it's purely for holding a "full file"?

It again comes back to the idea that maybe the loader shouldn't be using a `FullFileDataSet` to do its business, `FullFiIeDataSet` was originally made for the store.

yousefmoazzam: > I get it - so I think dataset.shape should be enough here?

I think this should work, using the block's shape to get the detector y/height length. In order for that to go wrong, the detector y/height dimension would have to be the slicing dim (so then the block wouldn't have the full previewed shape in the detetcor y dimension), and for the rotation wrapper we explicitly do this single sinogram gathering stuff under the assumption that it'll only ever be used in a "projection section".

httomo/loaders/standard_tomo_loader.py
yousefmoazzam: I ended up using a separate file to hold the types, I got a circular import trying to put the types into `__init__.py`, see [e2170d0](https://github.com/DiamondLightSource/httomo/pull/209/commits/e2170d0fcd7ae123a7cc062cfb77553bf93e6d3a)

httomo/method_wrappers/rotation.py
yousefmoazzam: Using the block shape rather than the base shape is done [1728181](https://github.com/DiamondLightSource/httomo/pull/209/commits/17281812846a5e0b831a335330d1a2e6f5a3ea20)

yousefmoazzam: After some discussion, we decided to address the `global_shape` concerns in another branch off of `feature/transparent-file-store`... :sweat_smile:

PR #214, opened 2024-02-16
Basic Monitoring / Profiling interface

team-gpu:
This PR adds the possibility to measure detailed performance metrics in an execution.

- It defines a `MonitoringInterface` protocol that has methods for reporting timings of various steps
- It has 2 implementations:
   - one for a summary on the console, which summarises some metrics and shows them in a human-readable way
   - and another which records each individual entry on its own and reports it in CSV format in the end
- A third implementation is an aggregator - it is a `MonitoringInterface` that can contain multiple other `MonitoringInterfaces`. It forwards any call to all of its children
- `TaskRunner` has been adapted to take an optional `MonitoringInterface` as a parameter, and if given, it measure various performance metrics and reports them to the monitor
- the CLI has 2 additional parameters to specify monitors to use (or not) + a filename in case it shouldn't go to the console
- the times are measured with context managers for CPU and GPU time

Further, it discovered some unnecessary transfers and efficiency issues with `calc_stats` and `save_intermediate_data`, which have been fixed directly.

The summary monitor produces an output like this:
```
Summary Statistics (aggregated across 1 processes):
  Total methods CPU time:     25.447s
  Total methods GPU time:      2.650s
  Total host2device time:      0.525s
  Total device2host time:      0.776s
  Total sources time    :      0.079s
  Total sinks time      :      1.028s
  Other overheads       :      1.092s
  ---------------------------------------
  Total pipeline time   :     27.646s
  Total wall time       :     27.646s
  ---------------------------------------
Method breakdowns:
                  find_center_vo :      0.679s ( 2.5%)
                remove_outlier3d :      0.318s ( 1.1%)
                       normalize :      0.038s ( 0.1%)
           paganin_filter_tomopy :      0.154s ( 0.6%)
     remove_stripe_based_sorting :      1.242s ( 4.5%)
                             FBP :      0.690s ( 2.5%)
          save_intermediate_data :      1.114s ( 4.0%)
                 median_filter3d :      0.334s ( 1.2%)
                 calculate_stats :      0.013s ( 0.0%)
                  save_to_images :     20.865s (75.5%)
```


httomo/monitors/__init__.py
```diff
@@ -0,0 +1,23 @@
...
+}
+
+def make_monitors(monitor_descriptors: List[str]) -> Optional[MonitoringInterface]:
+    if len(monitor_descriptors) == 0:
+        return None
+    
+    monitors: List[MonitoringInterface] = []
+    for descriptor in monitor_descriptors:
+        if descriptor not in MONITORS_MAP:
+            raise ValueError(f"Unknown monitor '{descriptor}'. Please choose one of {MONITORS_MAP.keys}")
```
yousefmoazzam, line 20: Nitpick, but I think `MONITORS_MAP.keys` is a method, so it needs to be executed by adding the parentheses:
```suggestion
            raise ValueError(f"Unknown monitor '{descriptor}'. Please choose one of {MONITORS_MAP.keys()}")
```

yousefmoazzam (approved):

PR #217, opened 2024-02-21
Use `data_offset` rather than `global_index` in loader

yousefmoazzam:
Provides a workaround to the following error during pipeline execution (for example, `test_run_diad_pipeline_gpu` exhibits the error):
```
  File "/httomo/httomo/runner/dataset.py", line 163, in _set_data
    self._global_index[dim] == 0
AssertionError: shape in slicing dim changed, which is not allowed
```

### DIFF
--- a/httomo/runner/loader.py
+++ b/httomo/runner/loader.py
@@ -136,6 +136,7 @@ class StandardTomoLoader(DataSetSource):
             global_index=self._chunk_index,
             chunk_shape=self._chunk_shape,
             shape=self._global_shape,
+            data_offset=(self._data_indices[0], 0, 0),
         )
 
         weakref.finalize(self, self.finalize)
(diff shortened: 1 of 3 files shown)


httomo/runner/loader.py
```diff
@@ -133,9 +134,12 @@ def __init__(
             angles=angles_arr,
             flats=flats_arr,
             darks=darks_arr,
-            global_index=self._chunk_index,
+            # Hardcoded for every MPI process because it's not being used in `FullFileDataSet`
+            # to offset block reading, that's what `data_offset` is being used for
+            global_index=(0, 0, 0),
```
team-gpu, line 139: I'm not sure about that actually - shouldn't the `global_index` be the index of the chunk within the global data, i.e. non-zero for processes of rank 1 and above, and the `data_offset` be `(self._data_indices[0], 0, 0)` ?

How would `block.global_index` look like for process 1, after it calls `make_block(0, 0, 10)`? That block should have a non-zero `global_index`.

team-gpu (changes requested):

team-gpu (approved):

PR #218, opened 2024-02-21
CPU memory limit parameter

team-gpu:
This adds a parameter to limit the CPU memory to be used by the pipeline. 

### How it works

- the `--max-memory` flag on the `httomo` command line can be used to specify the maximum amount of memory the pipeline can use
- internally we assume that half of that can be consumed the the dataset itself, as there are various overheads and we need space for inputs and outputs, reslices, etc.
- `TaskRunner` takes an optional parameter for the memory limit in bytes, and it uses that to limit the `available_memory`  used in max slices estimation. 
- `TaskRunner` also passes it on to the `DataSetStoreWriter` instances that it creates, so that these can make sure that the RAM-based store does not exceed this amount of memory
- `DataSetStoreWriter` throws `MemoryError` manually when the requested data is bigger than this limit, which will cause it to revert to file-based storage in exactly the same fashion as if physical memory was exhausted.

### Additional Bugfix

In addition, it fixes a small bug in `loader` - it was using the `FullFileDataSet.data` property to determine the data type, which in turn tried to read the whole chunk into memory at once. This fails for large datasets. Therefore we now extract the data type in the loader's constructor and simply return it from the property. And we make sure that `FullFileDataSet.data` is never called - all methods should work with blocks.

### Testing

The changes are fully covered with tests, and I confirmed with a 40GB dataset in the cluster that the limit is working and the code becomes runnable.


tests/runner/test_task_runner.py
```diff
@@ -141,6 +141,54 @@ def test_can_determine_max_slices_with_gpu_estimator(
         )
 
 
+@pytest.mark.parametrize("limit", [0, 10000])
+def test_can_determine_max_slices_with_gpu_estimator_and_cpu_limit(
```
yousefmoazzam, line 145: I'm trying to make sure I understand what this test is intended for. This doesn't seem to make any assertion about the actual max slices value that was determined by `TaskRunner.determine_max_slices()` for the single section, so I'm guessing it's not for making sure that the max slices are calculated correctly when constrained by the CPU memory limit.

I can see that there's a few mocked methods on the method wrapper objects. Is this test for checking that the CPU memory limit value defined in the `limit` variable is passed to each of the method wrapper's `calculate_max_slices()` method?

team-gpu: yes, that's correct. It checks that the limit passed in is actually limiting the available memory that is passed eventually down the `calc_max_slices` method. If a limit > 0 is given, that should be the limit that the slice estimator gets to see as available memory - not the `1e7` of total physical GPU memory.

yousefmoazzam: Ahhh I see. For the case when the CPU memory limit is lower than the GPU memory, if it can be proven that the `calculate_max_slices()` methods see the CPU memory limit rather than the GPU memory, then (assuming that the `TaskRunner.determine_max_slices()` method is working, which is the job of another test), then it can be assumed that the calculation of the max slices of the section is going to be done based on the CPU memory limit.

Yeah, I understand now that it wouldn't make sense to be asserting something about the result of `TaskRunner.determine_max_slices()` in this test, and that what the test checks for is sufficient to ensure expected behaviour of the CPU memory limit being passed, thanks for the explanation! :slightly_smiling_face:

yousefmoazzam (approved):

PR #219, opened 2024-02-23
Changes to memory estimation, d/f separate method

dkazanc:
@team-gpu : 
This is a bugfix PR to fix the issue that memory / slices estimation did not take into account the size of the darks and flats, and the fact that they are transferred lazily to the GPU on demand. The fix works as follows.

### `DataSet` properties

- `DataSet` and classes inheriting from it have new properties `has_gpu_flats` and `has_gpu_darks`, which return a boolean depending if the lazy GPU transfer has already happend and the data is present in GPU memory already

### `calc_max_slices` in the `TaskRunner`

- The memory estimation in `TaskRunner` first checks if any methods in the section need `darks` and `flats` as inputs, and if they are GPU methods
- If yes, it checks if the `darks` / `flats` have already been transferred to GPU (using `has_gpu_darks` and `has_gpu_flats`)
    - If yes, the figure obtained for `available_memory_gpu` is left as it was
    - If not, we know that they are transferred somewhere in this section, so the available memory is reduced by the amount taken for the darks and flats

### Memory estimator functions

- the `calc_max_slices` module type estimator functions now get extra keyword parameters for the darks and flats shape and dtype
- They can use this information if they need to, in case the `darks` and `flats` affect the estimate
- this was needed for the `remove_outlier3d` estimator described below

### Memory estimator for `normalize`

- Previously, multipliers with the direct method had been used for `darks` and `flats` in normalize
- This is however wrong - if these fields are already on the GPU, no new memory will be needed for them (normalise does not modify `darks` and `flats` in any way)
- The estimator has been modified to use `module` (since normalize can change the data type of `data`), and darks/flats are simply ignored there (no allocations happen in `normalize` depending on `darks` and `flats`)

### Memory estimator for `remove_outlier3d`

- This method is in fact replaced by the `dezinging` method wrapper in `httomo`
- This wrapper applies the method 3 times - to `data`, `darks`, and `flats`
- It does not modify the data type of the input though - that's left the same
- Therefore the estimator has been written as `module`, taking all this into account + a test has been added for the memory hook
- Note that because of the cupy memory pool (with memory re-used between calls in some cases), the hook test only works accurately if the data is large (which is what we care about in practice)


httomo/data/dataset_store.py
```diff
@@ -340,6 +340,10 @@ def aux_data(self) -> AuxiliaryData:
     def dtype(self) -> np.dtype:
         return self._data.dtype
 
+    @property
+    def has_gpu_flats(self) -> bool:
+        return self._data.has_gpu_flats
+
```
yousefmoazzam, line 346: Is this property needed anymore with the latest changes from `feature/transparent-file-store`, it doesn't seem to be used anywhere? The PR description mentions that this was a property added to `DataSet`, but now that `DataSet` is gone, does the store need it anymore?

httomo/runner/dataset_store_interfaces.py
```diff
@@ -80,6 +80,11 @@ def slicing_dim(self) -> Literal[0, 1, 2]:
     def aux_data(self) -> AuxiliaryData:
         """Auxiliary data"""
         ...  # pragma: no cover
+        
+    @property
+    def has_gpu_flats(self) -> bool:
+        """Check if flats array has already been cached on the GPU"""
+        ...  # pragma: no cover
```
yousefmoazzam, line 87: Same question as the other comment, is this needed still?

tests/method_wrappers/test_data_reducer.py
```diff
@@ -0,0 +1,73 @@
...
+        global_shape=GLOBAL_SHAPE,
+    )
+    newblock1 = wrp.execute(block1)
+
+    # we double flats and darks only, so we expect them to be twice the input
+    np.testing.assert_array_equal(newblock1.data, 1)
+    np.testing.assert_array_equal(newblock1.flats, 6)
+    np.testing.assert_array_equal(newblock1.darks, 8)
+
+    # it should only update the darks/flats once
```
yousefmoazzam, line 52: What this comment says is important, that the darks/flats should only be updated when the first block is processed, but I think this isn't being checked in the test?

I wonder if it may be useful to have another test for the data reducer wrapper which asserts this somehow? Ie, another test that somehow explicitly checks that when you process the second block, the darks/flats have definitely not been reassigned by the wrapper's `execute()` method?

yousefmoazzam: There's also the option of creating a mock of the `AuxiliaryData` object that is passed to the `DataSetBlock` constructor, to then "spy" on if the darks/flats setter get called only once, like what is done in one of the task runner tests to check that side outputs are only updated once https://github.com/DiamondLightSource/httomo/blob/0bff94bf70a75c9fadbfd14cae24e62a6c66d9fd/tests/runner/test_task_runner.py#L221

httomo/data/dataset_store.py
dkazanc: good point, even mypy marked it...

tests/method_wrappers/test_data_reducer.py
```diff
@@ -0,0 +1,73 @@
...
+    block2 = DataSetBlock(
+        data[2:2, :, :],
```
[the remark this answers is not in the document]
dkazanc: it would have made sense, yes. I guess it is a bug, comes from the dezinger test :)

tests/method_wrappers/test_data_reducer.py
dkazanc: Fair point indeed. I've added a couple of things to cover it, but feel free to change if you think there is a better way. 

- I'm checking that `_flats_darks_processed` variable set to `True` after executing the first block
- For the second block I change the auxiliary data for flats and darks, so if on the second block we do process f/d  again, then those updated arrays will be used instead of the initial ones. The assertion to initial f/d arrays proves that we ignore the updated ones but not executing the condition. Hope that this logic make sense?

yousefmoazzam: Thanks for the update!

Sorry, I'm admittedly a bit confused about the second auxiliary data object :sweat_smile:  I'm trying to make sure I understand how the extra `AuxiliaryData` object created in the test is being used, can I walk through what I _think_ I understand, and then you can correct me where I've got something wrong:
- `block1` is created with auxiliary data such that its darks is an array of 4's, and its flats is an array of 3's
- when `block1` is processed, the darks and flats should be doubled, so `newblock1` should have its darks as an array of 8's, and its flats as an array of 6's
- a new auxiliary data object is created with darks all 8's and flats all 6's, which are the same _values_ as the processed darks/flats in `newblock1`, but different array objects (ie, the values in the arrays are equal, but the array objects are not equal)
- `block2` is given this new auxiliary data that has darks all 8's and flats all 6's, so it has darks/flats which are the same values as the darks/flats from `newblock1`
- when `block2` is processed, its darks and flats should _not_ be doubled, so darks should stay as all 8's and flats should stay as all 6's
- at the end, the darks/flats of `newblock2` are asserted to be 8's for darks and 6's for flats

One thing that I might be misunderstanding is, because `block2` is given the same values for darks/flats as the doubled darks/flats from `newblock1`, how are we able to distinguish between:
- `newblock2` has 8's for darks and 6's for flats because it was given the new auxiliary data object
- `newblock2` has 8's for darks and 6's for flats because it has the updated darks/flats from `newblock1`

After all, `np.testing.assert_array_equal()` is only checking equality of the values in the arrays, it's not checking the array objects are the same object?

Is the idea that the darks/flats given to `block2` _haven't_ doubled from 8 to 18 and from 6 to 12 respectively?

yousefmoazzam: Something else that I'm not sure about is that different blocks should never be given different `AuxiliaryData` objects, a fundamental part of `AuxiliaryData`'s purpose was that every block should hold a reference to the _same_ `AuxiliaryData` object.

Creating two auxiliary data objects to give to two different blocks isn't something that should happen during pipeline execution, so it doesn't feel ideal to be testing a method wrapper under that condition, a method wrapper would always be given blocks which have a reference to the same `AuxiliaryData` object.

yousefmoazzam: Regarding the spy that was added, sorry, I think I didn't explain what I meant clearly enough when I suggested it as an option!

What I was meaning was that we could spy on the darks/flats setter on the `AuxiliaryData` object, to verify that the setters only get called once (ie, to check that the processing of the second block doesn't call the setters).

As it is now, the spy is looking at the wrapper's `execute()` method, which I don't think is doing any verification of the darks/flats being modified or not. What the spy is being used right now for is simply confirming that the wrapper's `execute()` method is called once after the test runs the `execute()` method, which isn't doing much -> we know that the `execute()` method has been called once because the test explicitly does this. I think what we're more interested in is spying on the call to any function that modifies the darks/flats (because checking how many times a function that modifies darks/flats is called will tell us how many times the darks/flats have been modified).

Does that make a bit more sense?

yousefmoazzam: I've made a few changes to the data reducer wrapper test in [494d2bf](https://github.com/DiamondLightSource/httomo/pull/219/commits/494d2bf04dccfa601469ed70e82df63124f857f6) to better illustrate what I was trying to describe.

(We don't have to keep that commit by any means, if it doesn't look useful we can revert locally and then force push to undo it, I just thought it was easier to show what I was meaning in code rather then describing it only in words :slightly_smiling_face:)

PR #239, opened 2024-03-12
fixes to work with median/dezinger 2d function, tests fixed etc

dkazanc:


httomo/method_wrappers/dezinging.py
```diff
@@ -30,18 +30,30 @@ def __init__(
...
+            method_name == "remove_outlier"
+        ), "Only remove_outlier is supported at the moment"
         self._flats_darks_processed = False
 
     def execute(self, block: DataSetBlock) -> DataSetBlock:
         self._gpu_time_info = GpuTimeInfo()
         # check if data needs to be transfered host <-> device
         block = self._transfer_data(block)
 
+        args = self._build_kwargs(self._transform_params(self._config_params), block)
```
dkazanc, line 51: @yousefmoazzam  I was trying a couple of ways to make it work with the test for wrapper which doesn't know anything about the parameters of the methods (like 'axis'). May be you know a better way rather than try/pass ?

yousefmoazzam: I'd have to dig deeper to understand more, but at a glance, what is the motivation for modifying `self._config_params`? Isn't the returned `args` value the "updated" params compared to `self._config_params`?

Can the more up-to-date `**args` be passed to `self.method()` instead of `**self._config_params`, like what is done in `images.py` https://github.com/DiamondLightSource/httomo/blob/4e08509907e0afc8a3ed588341d2970afffe4f59/httomo/method_wrappers/images.py#L65-L72

to then avoid the `try`/`except` entirely?

dkazanc: thanks, sorry should given more info. Yes the idea is to sift through `_transform_params` so that the `axis` auto parameter will be updated to an integer (depending on the `pattern`). 
You could indeed pass args directly, so this would work 
`block.data = self.method(**args)`
But because we also have darks and flats we need to take care of that somehow? Basically passing args but replacing data with d or f respectively. That is why I decided on simply getting the updated axis parameter from the args and feed it to `_config_params`.

httomo/method_wrappers/dezinging.py
```diff
@@ -30,18 +30,30 @@ def __init__(
...
+        try:
+            self._config_params["axis"] = args["axis"]
+        except:
+            pass
+
```
yousefmoazzam, line 56: Sorry, just to be clearer, I was suggesting something like the following:
```python
        args = self._build_kwargs(self._transform_params(self._config_params), block)
 
         with catch_gputime() as t:
            block.data = self.method(block.data, **args)
            if not self._flats_darks_processed:
                block.darks = self.method(block.darks, **args)
                block.flats = self.method(block.flats, **args)
                self._flats_darks_processed = True
```

where the code inside `with` block below has been updated to pass `**args` to all executions of `self.method()` in place of `**self._config_params`, so then projections/daks/flats are processed with the same updated parameters in `args`?

httomo/method_wrappers/dezinging.py
dkazanc: ah sorry, that won't probably work as args will contain data, i.e. `args['data']` will be pointing to a block of data, so `self.method` execute will fail for two data arguments being passed (I tried that).

yousefmoazzam: Ahhh I see, so `self._config_params` doesn't have a `data` key in it, but `args` has the transformed `axis` parameter value in it, and we need keys within both dicts... It's a bit fiddly huh!

yousefmoazzam: Hmm, if `axis` has a reasonable default value that we can pass in the case when `axis` isn't provided in the pipeline config, maybe we could do something like:
```python
self._config_params["axis"] = args.get("axis", DEFAULT_VALUE)
```

where we substitute the appropriate default value. That way, we get the `xis` parameter value if it's specified, but if it's not specified, then we use a sensible default value?

yousefmoazzam (approved):

PR #250, opened 2024-03-15
Dezing darks flats separately

### LINKED ISSUE #245: `DezingingWrapper` keeping projections in GPU memory while transferring darks/flats to GPU memory causes CUDA OOM
On GPUs with small memory, but large enough to hold all darks + all flats, but not always at the same time as holding a block in GPU memory (depending on the block size), the `remove_outlier` method fails with a CUDA OOM error.

With 20GB data that has:
- 20 darks (~0.2GB)
- 50 flats (~0.5GB)
- 1800 projections (~18GB)

`pc0074` (which has a GPU with 2GB of memory), is able to hold all darks and all flats in GPU memory. But, if the block size doesn't take the size of the darks/flats into account, then the block will be made too large to be able to fit all the darks and all the flats in GPU memory along with the block.

In `feature/transparent-file-store`, the state of the `DezingingWrapper` is such that it keeps a block, all darks, and all flats, in GPU memory before execution is returned to the task runner (which would transfer data to CPU when it needs to be written to the data store):
https://github.com/DiamondLightSource/httomo/blob/d9afc1f04e29a375cee17d33033d75f47a99da34/httomo/method_wrappers/dezinging.py#L57-L62

This seems to not agree with what the memory estimation for `remove_outlier` now is (changes were recently done to memory estimation of `remove_outlier` in #239), where accounting for darks/flats looks to not be present: https://github.com/DiamondLightSource/httomo/blob/d9afc1f04e29a375cee17d33033d75f47a99da34/httomo/methods_database/packages/external/httomolibgpu/1.2/httomolibgpu.yaml#L12-L20

It seems to be the case that either:
1. the memory estimation for `remove_outlier` needs to account for darks/flats
2. projections need to be transferred to CPU _before_ darks are transferred to GPU, and the darks need to be transferred to CPU (after processing) _before_ the flats are transferred to the GPU

yousefmoazzam:
An experimental attempt to fix #245

### DIFF
--- a/httomo/method_wrappers/dezinging.py
+++ b/httomo/method_wrappers/dezinging.py
@@ -55,39 +55,39 @@ class DezingingWrapper(GenericMethodWrapper):
         # tested in the yaml checker potentially
         self._config_params["axis"] = args.get("axis", 0)
 
-        with catch_gputime() as t:
-            block.data = self.method(block.data, **self._config_params)
-            if self.is_gpu:
+        if self.is_gpu:
+            with catch_gputime() as t:
+                block.data = self.method(block.data, **self._config_params)
                 block.to_cpu()
                 gpumem_cleanup()
 
-            if not self._flats_darks_processed:
-                if self.is_gpu:
+                if not self._flats_darks_processed:
                     block.aux_data.set_darks(
                         xp.asnumpy(
                             self.method(
-                                block.aux_data.get_darks(gpu=True), **self._config_params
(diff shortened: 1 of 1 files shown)


yousefmoazzam:
Closing for now, as an alternative has been explored with b4cb013

dkazanc:
closing this branch as the solution was to "flatten" flats and darks prior being passed to dezinger and normaliser.

Issue #299, opened 2024-04-29
autogeneration of yaml templates with the docs build


dkazanc:
We need to autogenerate the yaml templates during the documentation build instead of referring to the directory where all yaml templates stored. The latter should be deleted with templates automatically uploaded to docs. 
First make it work locally then deal with the Actions build. 

- [x] sphinx-build script must generate yaml templates for specific (supported) versions of back-ends given in the doc-conda-requirements.yml. This means that the correct version of the package must be on the conda cloud already.  The issue https://github.com/DiamondLightSource/httomo/issues/285 should be retired.
- [x] Ensure that Actions do what they supposed to and upload the correct templates to documentation.
- [x] Delete versions.yaml
Before delete `yaml_templates` from the repo one needs to resolve https://github.com/DiamondLightSource/httomo/issues/300

PR #307, opened 2024-05-02
Auto generator of yamls for backends during documentation build

dkazanc:
- Mostly changing GA recipe to accommodate auto generation of yaml templates during the build of documentation. 
- So far `yaml_templates` folder contains yamls but they are not used in the GA build. They are required mostly for yaml checker AFAIAA, see more in  #299


yousefmoazzam:
Whenever this is merged, before that's done, would it be ok to tidy up the commit history? Some commits are doing the opposite of each other, like b971296 and 04439bb, so I think there's plenty of commits that can either be removed entirely, or groups of commits that could be squashed together.

I do understand that all these commits need to be made when testing github actions, I think it's fine (and basically necessary with CI :sweat_smile:...) for WIP stuff, but cleaning things up once it's all working would be great :slightly_smiling_face:

Issue #331, opened 2024-05-14
Nonexistent method in pipeline YAML causes YAML checker to fail unexpectedly


yousefmoazzam:
Suppose the following pipeline YAML file is given to the YAML checker (this is a minimal example that reproduces the issue):
```yaml
- method: standard_tomo
  module_path: httomo.data.hdf.loaders
  parameters:
    name: tomo
    data_path: entry1/tomo_entry/data/data
    image_key_path: entry1/tomo_entry/instrument/detector/image_key
    rotation_angles:
      data_path: /entry1/tomo_entry/data/rotation_angle
    dimension: 1
    pad: 0
- method: find_center_vol
  module_path: httomolibgpu.recon.rotation
  parameters:
    ind: mid
    smin: -50
    smax: 50
    srad: 6.0
    step: 0.25
    ratio: 0.5
    drop: 20
```

and notice the incorrect method name `find_center_vol` (should be `find_center_vo`).

The YAML checker correctly reports this as being something incorrect in the YAML. However, afterwards, the YAML checker then fails abruptly:
```
Checking that the YAML_CONFIG is properly indented and has valid mappings and tags...
Sanity check of the YAML_CONFIG was successfully done...

Checking that the first method in the pipeline is a loader...
Loader check successful!!

'httomolibgpu.recon.rotation/find_center_vol' is not a valid method. Please recheck the yaml file.
Traceback (most recent call last):
  File "/dls/science/users/twi18192/conda-envs/httomo/lib/python3.10/runpy.py", line 196, in _run_module_as_main
    return _run_code(code, main_globals, None,
  File "/dls/science/users/twi18192/conda-envs/httomo/lib/python3.10/runpy.py", line 86, in _run_code
    exec(code, run_globals)
  File "/dls/science/users/twi18192/httomo/httomo/__main__.py", line 4, in <module>
    main()
  File "/dls/science/users/twi18192/conda-envs/httomo/lib/python3.10/site-packages/click/core.py", line 1157, in __call__
    return self.main(*args, **kwargs)
  File "/dls/science/users/twi18192/conda-envs/httomo/lib/python3.10/site-packages/click/core.py", line 1078, in main
    rv = self.invoke(ctx)
  File "/dls/science/users/twi18192/conda-envs/httomo/lib/python3.10/site-packages/click/core.py", line 1688, in invoke
    return _process_result(sub_ctx.command.invoke(sub_ctx))
[...]

Issue #321, opened 2024-05-08
Uncomment commented-out CLI tests?


yousefmoazzam:
There are several commented-out CLI tests: https://github.com/DiamondLightSource/httomo/blob/833936aaa1049feac5c55a2195e99454dee85fcc/tests/test_cli.py#L26-L67

1. `test_cli_noargs_raises_error` is asserting that the CLI raises an error if no args at all are passed
2. `test_cli_check_pass_data_file` is asserting that the YAML checker produces output saying it's checking paths in the input data if it's given a data file
3. `test_cli_pass_output_folder` is asserting that the `user.log` file is found in the custom output dir specified (this flags is reintroduced in #306)

I'm not sure why 1 was commented out?

With the YAML checker stuff merged in #201, it seems like the YAML check test in 2 should be uncommented.

When #306 is merged, I think 3 should be uncommented.

yousefmoazzam:
Completed in merge of #332

PR #332, opened 2024-05-14
Fixing tests

dkazanc:
- [ ] uncommenting tests in #321 
- [ ] Introducing fixes in #331

### DIFF
--- a/tests/test_cli.py
+++ b/tests/test_cli.py
@@ -23,17 +23,15 @@ def test_cli_help_shows_help():
         .startswith("Usage: python -m httomo")
     )
 
-"""
-# TODO possibly re-enable after yaml checker is complete
+
 def test_cli_noargs_raises_error():
     cmd = [sys.executable, "-m", "httomo"]
     try:
         subprocess.check_output(cmd)
     except subprocess.CalledProcessError as e:
         assert e.returncode == 2
-"""
 
-"""
+
 def test_cli_check_pass_data_file(standard_loader, standard_data):
     cmd = [sys.executable, "-m", "httomo", "check", standard_loader, standard_data]
     check_data_str = (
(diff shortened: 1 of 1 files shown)


tests/test_cli.py
```diff
@@ -23,27 +23,23 @@ def test_cli_help_shows_help():
...
     cmd = [sys.executable, "-m", "httomo", "check", standard_loader, standard_data]
     check_data_str = (
         "Checking that the paths to the data and keys in the YAML_CONFIG file "
         "match the paths and keys in the input file (IN_DATA)..."
     )
     assert check_data_str in subprocess.check_output(cmd).decode().strip()
-"""
 
-"""
 def test_cli_pass_output_folder(
```
yousefmoazzam, line 43: Ah, in #306, the equivalent of `test_cli_pass_output_folder()` was added back in the form of `test_cli_pass_create_folder()`, I forgot that I could have uncommented this one instead... :sweat_smile:

Could you remove one to avoid duplicating the test? Sorry for the extra work, that's my fault for not using the commented-out one.

yousefmoazzam (changes requested):
Cool, thanks for addressing the three commented-out tests!

I think that the two issues in the original post are unrelated, #331 isn't doing anything with tests whereas #321 is, so they would make sense to be dealt with in separate PRs.

So, I reckon that once the duplicated test has been removed that I mention below, this can be merged, and #331 can be dealt with elsewhere :slightly_smiling_face:

dkazanc:
@yousefmoazzam  OK with that?

yousefmoazzam (approved):
Looks ok to me!

I guess the YAML validation print change is tiny and doesn't change any functionality, so it's not the worst thing to be in here. For the future, anything that is bigger than that and is unrelated to the PR's intention is better put in a separate PR for the sake of making it easier to understand which changes were introduced where :slightly_smiling_face: 

In this case, the only downside is simply that it's confusing/odd to see a PR titled "Fixing tests" changing a print statement that is completely unrelated to any tests. And even then with little downside, from an organisation point of view it feels "tidier" to keep things compartmentalised, but I do appreciate this is getting to more subjective territory.

In other situations though, the downside could be much larger, such as having multiple different functionalities being changes within the same PR. This makes reviewing harder compared to if they were split among different PRs, due to the cognitive load of keeping multiple functionalities in mind when reviewing.

Issue #300, opened 2024-04-30
Generate YAML templates during installation


yousefmoazzam:
The YAML checker depends on the YAML templates existing in order to perform the checks that the methods specified in a pipeline are valid methods in the associated packages: https://github.com/DiamondLightSource/httomo/blob/f53709ac6350898dcfe874961241d98e3a722f84/httomo/yaml_checker.py#L130-L146

Thus, it would make sense to generate the YAML templates during installation of httomo, so then the YAML checker is ready to use after installation.

yousefmoazzam:
A few comments:
- if it's difficult to do this with `pyproject.toml` alone, _and_ poetry plays nicely with conda, then using poetry would be ok
- if it's doable without poetry, that would be nice, since it would be another thing to deal with for packaging
- I've not heard this notion before of generating files for an installation during test execution, does this involve running `pytest` during the `pip install`/`conda install` somehow?  (I do recall there being the ability to run tests during `conda build`, but that's all I remember). What are the benefits of having the YAML template generation wrapped within tests, is it a means of detecting failure of the YAML templates generation?

yousefmoazzam:
Yeahhh, this is looks quite tricky to accomplish...

With some fiddling, it was possible to provide a custom step in the build process that occurs during a `pip install` to run the YAML templates generator script for the various backends in the `generate-yaml-templates` branch https://github.com/DiamondLightSource/httomo/blob/38f3ac81ef01126243cccfe7c763ceb0cf71d72c/setup.py

This does indeed manage to generate the `yaml_templates/` dir in the `build/` dir that eventually has its contents copied to the installation destination by `pip install`

However, when it gets to calling the function in `yaml_templates_generator.py`, the issue is that it needs to import the various backend packages in order to inspect the signatures of the functions... The backends aren't installed in the build environment that `pip` creates, and thus the YAML templates generation fails at this point:
```
Traceback (most recent call last):
    File "/dls/science/users/twi18192/conda-envs/httomo-install-test/lib/python3.10/site-packages/pip/_vendor/pyproject_hooks/_in_process/_in_process.py", line 353, in <module>
      main()
    File "/dls/science/users/twi18192/conda-envs/httomo-install-test/lib/python3.10/site-packages/pip/_vendor/pyproject_hooks/_in_process/_in_process.py", line 335, in main
      json_out['return_val'] = hook(**hook_input['kwargs'])
[...]

PR #341, opened 2024-05-21
Move YAML templates dir into `httomo/` subdir

### LINKED ISSUE #335: Fix path to YAML templates in YAML checker when YAML templates are eventually copied during installation
Regardless of if/when #300 is done, there is still something else to deal with.

Assume the root directory for the httomo repo being used for the source for an installation is `/scratch/user/httomo`, and that the root directory for httomo when installed in a conda env is `/dls_sw/apps/httomo/2.0.0rc1/lib/python3.10/site-packages/httomo`

The YAML checker file `yaml_checker.py` is in the subdir `/scratch/user/httomo/httomo/yaml_checker.py`, and the YAML templates when in version control in the repo are in the root directory `/scratch/user/httomo/yaml_templates`.

When running httomo locally in dev mode, this allows the following code to correctly find the YAML templates dir relative to the `yaml_checker.py` file: https://github.com/DiamondLightSource/httomo/blob/8b77eb0776b1ae02d4832e5878752051fce1213d/httomo/yaml_checker.py#L347

However, when httomo is installed via `pip install` or `conda install` into `/dls_sw`, the root directory stuff like `/scratch/users/httomo/pyproject.toml` is not copied to the conda env, only the contents of `/scratch/user/httomo/httomo` subdir is copied. Hence, the dir for httomo in the conda env has only the contents of `/scratch/user/httomo/httomo` in `/dls_sw/apps/httomo/2.0.0rc1/lib/python3.10/site-packages/httomo`.

This means that the only place to copy the YAML templates would be `/dls_sw/apps/httomo/2.0.0rc1/lib/python3.10/site-packages/httomo/yaml_templates`.

However, notice in that situation, `yaml_checker.py` are now adjacent to each other rather than `yaml_checker.py` being in a dir "below" the YAML templates dir.

Ie, compare:
```
/dls_sw/apps/httomo/2.0.0rc1/lib/python3.10/site-packages/httomo/yaml_checker.py
/dls_sw/apps/httomo/2.0.0rc1/lib/python3.10/site-packages/httomo/yaml_templates
```

to
```
/scratch/user/httomo/httomo/yaml_checker.py
/scratch/user/httomo/yaml_templates
```

and it becomes clearer that their locations relative to each other will change if the YAML templates are copied to the "root httomo dir" in the conda env.

This essentially requires the path of the YAML templates dir in the YAML checker to be changed to:
```python
parent_dir = Path(__file__).parent
```

and is what was done to hackily install httomo in `/dls_sw` to make it work

yousefmoazzam:
Fixes #335

This is a tentative attempt to try and resolve #335, motivated by the slightly tedious need to generate the YAML templates for all backends + httomo manually after installing httomo in a conda env in `/dls_sw/apps/httomo`.

I agree that it's not great to have the YAML templates adjacent to the python source code (ie, it feels weird to have YAML files among python files). However, given the description in #335 of how the source files are copied into the conda env's dir, it seems like it's one of the only ways to do this. Plus, we already have YAML files among the python source code in the installation, the YAML files in the methods database are there.

Once #300 is done, then the `yaml_templates/` dir can go away completely from version control, which means the YAML templates need only be adjacent to the source code in the installation dir, which maybe isn't as nasty looking as if they're adjacent the source code during development?... :sweat_smile: 

Acceptance criteria checklist
- [x] All tests pass
- [ ] Consensus that it's a reasonable solution to the problem


dkazanc:
Thanks! I'm OK with it I think. There is no a specific need for it to exist outside the source code base. The only question is why to move the folder if #300 can be easier (?) implemented if the templates become a part of the source base? Or I guess the folder should exist in any case?

yousefmoazzam:
Yep, as you say, the directory `yaml_templates/` will have to be created in the installation directory (as httomo's YAML checker needs it at runtime), regardless of if it's in version control or not. This means that deciding where the best place to put it needs to happen regardless of #300.

As for #300 being easier to implement, I don't actually know how easy or difficult it is to do, I haven't worked much with the YAML template generation. If I recall correctly, originally I questioned why the YAML templates were under version control at all, and this then led to the fact that there's some difficulty in autogenerating the YAML templates. Which meant that they were manually written to some extent, and hence needed to be under version control?

yousefmoazzam:
If it is easier to do #300, then for sure, I'd suggest to favour doing that over this - having the YAML templates under version control did feel a bit messy.

If #300 is tricky though, then this could be considered a temporary solution that makes installation easier to deal with until autogeneration is possible.

dkazanc:
Ok, thanks it makes sense. Well I guess we can move in steps indeed. First move the folder that consists templates under source code base and then think about #300. I think the main difficulty there is to execute python YAML generator script during the build. There are some [hints](https://stackoverflow.com/questions/59286983/how-to-run-a-script-using-pyproject-toml-settings-and-poetry) to do that using poetry, but may be there are other solutions? I, myself, was thinking that may be the yaml generator functions should be a part of the tests that must run after the installation. Essentially, running not the script itself by Python functions. Could be nonsense of course...

yousefmoazzam:
Ok cool, that seems sensible: do this for now, and then do #300 as the final solution :slightly_smiling_face: 

Re: poetry + etc, let's put the discussion for fixes of #300 over there, to keep information related to its solution in one place.

PR #347, opened 2024-05-24
Generate YAML templates during conda package build

### LINKED ISSUE #300: Generate YAML templates during installation

yousefmoazzam:
Fixes #300

A tentative attempt to be able to provide the YAML templates at installation time.

The strategy here is to generate the YAML templates during the build of the conda package (by adding the relevant backend dependencies that are needed for the YAML templates generation), and place them in the httomo installation dir in the conda env when the conda package is installed.

Acceptance criteria checklist:
- [x] Other devs confirm that installation of locally built conda package (`httomo-2.0.0rc1-py39_0.tar.bz2`) installs YAML templates into `/path/to/conda/env/lib/site-packages/python3.9/httomo/yaml_templates/`


yousefmoazzam:
Some notes:
- when testing out locally, the `test` field in `meta.yaml` was commented out in order to make the build run faster
- the built httomo conda package was installed locally into a fresh conda env and was confirmed to have the YAML templates in the httomo installation dir

dkazanc:
Thanks @yousefmoazzam. I've managed to conda build the repo and then I checked the contents:
I see 3 locations with the templates. 2 are outside of htttomo and one `yaml_templates` inside httomo. Any reason why those 2 folders are outside httomo? `templates` is the older name for `yaml_templates` ? 
image

yousefmoazzam:
@dkazanc Yeah I've had similar issues in the past with old the templates dir being copied into the `site-packages/` dir.

By any chance, do you have a `build/lib/` dir in your httomo repo which you did the conda build from? And if so, does it happen to have a `build/lib/templates/` dir in it? If that's the case, I'd suggest to delete the entire `build/` dir and try doing the conda package build again.

From my experiences, the `build/` dir gets stuff in it generated when doing a `pip install` (which `conda build` also will do). However, files/dirs in it aren't deleted + recreated when doing multiple `pip install` or `conda build` executions. Meaning, old files/dirs from previous states of the repo can accumulate, and thus the old `templates/` dir and the old place for the `yaml_tenplates/` dir can be leftover from previous installations/builds.

dkazanc:
Great, thanks Yousef. I can confirm that this works for me now as expected. OK to merge?

yousefmoazzam:
Cool, glad that worked to get rid of those unwanted dirs!

Yep, I think this is good to merge, will do now.

PR #353, opened 2024-05-29
Build conda packages for openmpi "external" and "regular" build variants

### LINKED ISSUE #352: Generate conda packages for two different openmpi build strings
There are two main installation use cases of httomo, based on the differences between the hardware httomo is being run on:
1. Usage at DLS
An "external" openmpi conda package build is needed (to make httomo compatible with the OpenMPI built at DLS with SLURM support)

2. Usage outside of Diamond SLURM cluster usage
Any openmpi build will be fine (and probably other MPI implementations would be fine)

The `meta.yaml` file involved in the building of a conda package offers jinja templating as a means to generate multiple conda package builds https://docs.conda.io/projects/conda-build/en/latest/resources/define-metadata.html#templating-with-jinja, which likely could be used in some manner to achieve the above two conda package builds of httomo for easier packaging and installation.

yousefmoazzam:
Fixes #352
Depends on #347 being merged first

## Package naming

The generated package names look something like the following (fixing it for python 3.10, and httomo version v2.0.0rc1 for the example):
```
httomo-2.0.0rc1_openmpi_regular
httomo-2.0.0rc1_openmpi_external
```

"Regular" and "external" refer to the build strings of the `openmpi` package that the different httomo package builds are based on.

I chose these names as a starting point just to get the work done, but the names themselves can easily be changed, so opinions on better naming are certainly welcome :)

## Installation in practise with these changes

Assume that  the packages are uploaded to the httomo conda channel.

Installation of the package using the "external" openmpi build variant (for installation at DLS) would be done via:
```
mamba install -c https://conda.anaconda.org/httomo/ httomo=2.0.0rc1=*external*
```

Installation of the package using the  "regular" openmpi build variant (for installation outside of DLS) would be done via:
```
mamba install -c https://conda.anaconda.org/httomo/ httomo=2.0.0rc1=*regular*
```

## Acceptance criteria checklist
- [x] The built packages install the correct `openmpi` package build variant
- [x] There is a consensus on the names of the different build variants of the httomo package

### DIFF
--- a/conda/recipe/conda_build_config.yaml
+++ b/conda/recipe/conda_build_config.yaml
@@ -3,3 +3,6 @@ python:
   - 3.10
 numpy:
   - 1.24
+openmpi_build:
+  - regular
+  - external
--- a/conda/recipe/meta.yaml
+++ b/conda/recipe/meta.yaml
@@ -11,7 +11,7 @@ source:
 
 build:
   preserve_egg_dir: False
-  number: 0
+  string: py{{ py }}_openmpi_{{ openmpi_build }}
 
 test:
   files:
@@ -60,6 +60,11 @@ requirements:
     - pyyaml
     - numpy<=1.25
     - nvtx
+    {% if openmpi_build == 'regular' %}
+    - openmpi
+    {% else %}
+    - openmpi * *external*
+    {% endif %}
     - python
     - plumbum
     - tomobar


dkazanc:
thanks @yousefmoazzam ! What happens if you do?
`mamba install -c https://conda.anaconda.org/httomo/ httomo` ? Does it install the regular version? 
Sorry `external` is slightly confusing as from the Diamond standpoint the regular installation will be external. But I also do not want to fix any Diamond wording here as this installation might be used elsewhere.  With `external`, does that mean that openmpi is actually not installed at all?

yousefmoazzam:
Mmm, fair question about what build variant would be installed by leaving off any constraint on it and doing:
```
mamba install -c https://conda.anaconda.org/httomo/ httomo
```

I don't know what conda/mamba is doing to decide which build variant should be installed if both are ok to be installed into the existing conda env, that might require digging into the docs for how dependency resolution is done. But when I run that command locally with an env that has python 3.10 installed in it, mamba consistently picks up the "regular" build variant, ie, this one:
```
httomo-2.0.0rc1_openmpi_regular
```

to install, for whatever reason.

yousefmoazzam:
> Sorry external is slightly confusing as from the Diamond standpoint the regular installation will be external. But I also do not want to fix any Diamond wording here as this installation might be used elsewhere. With external, does that mean that openmpi is actually not installed at all?

Yep, the confusing naming of "external" is a very reasonable comment haha. It was one I also got confused with, and was why I mentioned I was happy to change the naming.

As things are currently, "external" means "installing the external openmpi build variant". The external openmpi build variant doesn't install openmpi, it instead installs some dummy package, so then the user can use their own local openmpi. This is what we would use at DLS.

So, rather confusingly, the installation of httomo would be:
- install the "external" httomo build variant at DLS
- install the "regular" httomo build variant elsewhere

dkazanc:
I guess I'm ok with `external`. May be `specified` would work or something that points that we need a specific openmpi dependency here?

yousefmoazzam:
I'm fine with any name to be honest. The reason being, I suspect that whatever we choose, we're going to have to explain it somewhere anyway (dev docs maybe), that there's this "external" build variant of openmpi that one of the httomo packages is built with.

So whatever name you're happy with @dkazanc for the build strings, I'm happy to go with too, just let me know and I can make the change accordingly.

dkazanc:
let's go with `external` :)

PR #361, opened 2024-06-05
FBP memory estimation corrections

dkazanc:
- [x] more accurate estimator and modified tests
- [x] test it with larger data on the node

### DIFF
--- a/httomo/methods_database/packages/external/httomolibgpu/supporting_funcs/recon/algorithm.py
+++ b/httomo/methods_database/packages/external/httomolibgpu/supporting_funcs/recon/algorithm.py
@@ -100,11 +100,16 @@ def _calc_memory_bytes_FBP(
         )
         / SLICES
     )
+    # The multiplier_heuristic comes from the empirical testing of the module for various
+    # data sizes. So far, the need for it is cannot be explained from the algorithmic
+    # standpoint. Also, there is no association of it with filtered_freq_slice array,
+    # as we just need to bump up the memory to avoid the OOM error.
+    multiplier_heuristic = 2
 
     tot_memory_bytes = int(
-        input_slice_size
+        2 * input_slice_size
         + filtered_input_slice_size
-        + filtered_freq_slice
+        + multiplier_heuristic * filtered_freq_slice
         + fftplan_size
         + ifftplan_size
         + recon_output_size
--- a/httomo/methods_database/packages/external/httomolibgpu/supporting_funcs/recon/algorithm.py
+++ b/httomo/methods_database/packages/external/httomolibgpu/supporting_funcs/recon/algorithm.py
@@ -23,7 +23,7 @@
 import math
 from typing import Tuple
 import numpy as np
-
+from httomo.cufft import CufftType, cufft_estimate_1d
 
 __all__ = [
     "_calc_memory_bytes_FBP",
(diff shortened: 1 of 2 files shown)


httomo/methods_database/packages/external/httomolibgpu/supporting_funcs/recon/algorithm.py
```diff
@@ -66,28 +66,54 @@ def _calc_memory_bytes_FBP(
...
+    )
 
     tot_memory_bytes = int(
-        2 * in_slice_size
-        + filtered_in_data
-        + freq_slice
+        input_slice_size
+        + filtered_input_slice_size
+        + filtered_freq_slice
         + fftplan_size
```
dkazanc, line 108: Testing with larger datasets (40gb as 80 doesn't work currently) I had to double the `fftplan_size` for it to pass. @yousefmoazzam do you know any reason what that should be the case or what is the current estimate missing? We know now that FFT filtering part is the most memory hungry one there, so we just looking at it. cheers

yousefmoazzam: I'll keep looking more closely, but on a first pass at the FBP code, something that sticks out to me is a question of if this line is being accounted for in the memory estimator https://github.com/dkazanc/ToMoBAR/blob/54137829b6326406e09f6ef9c95eb35c213838a7/tomobar/methodsDIR_CuPy.py#L135

The [cupy docs](https://docs.cupy.dev/en/stable/reference/generated/cupy.swapaxes.html) say that `swapaxes()` will return a view of the array. Non-trivial views of a multi-dimensional array; ie, anything that involves more than simply picking a subset with no fancy indexing will in general be rearranging the ordering of elements I think.

Swapping the order of two axes is a non-trivial operation as far as the arrangement of elements. So it'll probably rearrange elements in such a way that if the original element layout was C-contiguous, the layout of the elements in the view with the axes swapped won't be C-contiguous.

Forcing such a view of data to be C-contiguous feels like it'll most likely result in a copy, and in this case it's the entire data having its axes swapped and being forced to be C-contiguous.

dkazanc: Thanks, this might explain why Jorg (I think) did initial estimate by [doubling](https://github.com/DiamondLightSource/httomo/blob/a5067f08965a44697aa31d593ee0686ed00cfbd4/httomo/methods_database/packages/external/httomolibgpu/supporting_funcs/recon/algorithm.py#L85) the size of the input slice.  Will give it a test...

dkazanc: unfortunately even when doubling the `input_slice_size` there is not enough memory. And the OOM failure is inside  `filtersinc3D_cupy` part. [This](https://github.com/dkazanc/ToMoBAR/blob/54137829b6326406e09f6ef9c95eb35c213838a7/tomobar/fourier.py#L70) shouldn't create a copy?

yousefmoazzam: That's correct, that in-place multiplication shouldn't cause a copy of the array to be created.

It's tedious work, but at times when doing work on memory estimators before, I ended up using cupy's [`LineProfileHook`](https://docs.cupy.dev/en/stable/reference/generated/cupy.cuda.memory_hooks.LineProfileHook.html) to look at all the memory allocations being done that cupy is aware of.

It can be used like the hook used in the memory estimator tests, via `with`. That might be useful to do to see if there's any allocations happening in the cupy filtering that hasn't been accounted for.

PR #364, opened 2024-06-12
fix for cpu max memory estimation

dkazanc:
fixing the issue of `MAX_CPU_SLICES` not enabled for the all-cpu section.

### DIFF
--- a/httomo/runner/task_runner.py
+++ b/httomo/runner/task_runner.py
@@ -340,9 +340,4 @@ class TaskRunner:
             max_slices_methods[idx] = min(max_slices, slices_estimated)
             non_slice_dims_shape = output_dims
 
-        if not has_gpu:
-            section.max_slices = min(
-                min(max_slices_methods), httomo.globals.MAX_CPU_SLICES
-            )
-        else:
-            section.max_slices = min(max_slices_methods)
+        section.max_slices = min(max_slices_methods)
--- a/httomo/runner/task_runner.py
+++ b/httomo/runner/task_runner.py
@@ -7,9 +7,7 @@ import os
 import tqdm
 from mpi4py import MPI
 
-import httomo
-from httomo.globals import *
-
+import httomo.globals
 from httomo.data.dataset_store import DataSetStoreWriter
 from httomo.runner.method_wrapper import MethodWrapper
 from httomo.runner.block_split import BlockSplitter
@@ -192,7 +190,7 @@ class TaskRunner:
 
     def _prepare(self):
         self._log_pipeline(
-            f"See the full log file at: {run_out_dir}/user.log",
+            f"See the full log file at: {httomo.globals.run_out_dir}/user.log",
         )
         self._check_params_for_sweep()
         self._load_datasets()
@@ -305,7 +303,7 @@ class TaskRunner:
 
         # if section consists of all cpu method then MAX_CPU_SLICES defines the block size
         if not has_gpu:
-            section.max_slices = min(MAX_CPU_SLICES, max_slices)
+            section.max_slices = min(httomo.globals.MAX_CPU_SLICES, max_slices)
 
         nsl_dim_l = list(data_shape)
         nsl_dim_l.pop(slicing_dim)
@@ -344,6 +342,8 @@ class TaskRunner:
             non_slice_dims_shape = output_dims
 
         if not has_gpu:
-            section.max_slices = min(min(max_slices_methods), MAX_CPU_SLICES)
+            section.max_slices = min(
+                min(max_slices_methods), httomo.globals.MAX_CPU_SLICES
+            )
         else:
             section.max_slices = min(max_slices_methods)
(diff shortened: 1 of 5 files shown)


httomo/runner/task_runner.py
```diff
@@ -290,12 +292,20 @@ def _count_tuple_values(self, d: Dict[str, Any]) -> int:
...
+
+        # loop over all methods in section
+        has_gpu = False
+        for idx, m in enumerate(section):
+            if m.implementation in ["gpu", "gpu_cupy"]:
+                has_gpu = True
+
+        # if section consists of all cpu method then MAX_CPU_SLICES defines the block size
+        if not has_gpu:
+            section.max_slices = min(MAX_CPU_SLICES, max_slices)
```
yousefmoazzam, line 308: If this condition is encountered (ie, we have confirmed that all methods in the section are CPU), is there any need to carry on with the rest of the code in this function? Can we just `return` after taking the max slices to be the min of the `MAX_CPU_SLICES` and `max_slices`?

httomo/runner/task_runner.py
```diff
@@ -334,8 +344,6 @@ def determine_max_slices(self, section: Section, slicing_dim: int):
             non_slice_dims_shape = output_dims
 
         if not has_gpu:
-            section.max_slices = min(
-                min(max_slices_methods), httomo.globals.MAX_CPU_SLICES
-            )
+            section.max_slices = min(min(max_slices_methods), MAX_CPU_SLICES)
```
yousefmoazzam, line 347: I know that this is just the formatter changing things and no logic has been altered, but it did make me wonder why there is another check of `if not has_gpu`.

What I mean is, as the code stands, why is there two different places that the max slices gets set for when the section contains all CPU methods? It feels like there should be only one place that this occurs, it feels confusing that it gets set once earlier, and then once more here? Or have I misunderstood what this part of the code is doing?

tests/runner/test_task_runner.py
```diff
@@ -5,6 +5,7 @@
 import numpy as np
 from pytest_mock import MockerFixture
 import httomo
+from httomo.globals import *
```
yousefmoazzam, line 8: I'm admittedly not very fond of `*` imports, it's hard to tell what's being brought into scope. Among the formatter changes I can't immediately tell what is being used from `httomo.globals` or where it's being used, could you point me to what's needed? (My guess was `MAX_CPU_SLICES`, but I can't see it anywhere in the additions here?).

httomo/runner/task_runner.py
```diff
@@ -8,6 +8,8 @@
 from mpi4py import MPI
 
 import httomo
+from httomo.globals import *
```
yousefmoazzam, line 11: This is personal preference, so feel free to take it or leave it.

I tend to prefer avoiding `*` imports. I can see that `run_out_dir` and `MAX_CPU_SLICES` are coming from `httomo.globals`, but I personally prefer to know where they're coming from explicitly (as things were before), especially if they're global variables.

Seeing `httomo.globals.run_out_dir` makes it somewhat clear to me how the task runner has gotten ahold of the output dir, but seeing just `run_out_dir` makes me wonder how/why the task runner has the output dir value floating around.

yousefmoazzam: And actually, perhaps a more serious issue with this is that I'm unsure if the global variables imported via the `*` would be updated from the perspective of the task runner after the import is done. Ie, I'm not sure if `run_out_dir` and `MAX_CPU_SLICES` being updated wherever they get updated will be reflected in the task runner, does this work if a pipeline is executed? Does `run_out_dir` get seen as the value it's updated to in `cli.py`, or is it still `None` when the task runner starts doing things?

dkazanc: this change was actually because of the tests failing due to `httomo.globals` couldn't be located because of  `import httomo`. 
So I've changed to `import httomo.globals`, seems to work now.

httomo/runner/task_runner.py
dkazanc: good point, it also checked for `m.is_gpu` bellow so I added it in the condition,  in order to remove the bit bellow 
`if m.implementation in ["gpu", "gpu_cupy"] or m.is_gpu:`

httomo/runner/task_runner.py
dkazanc: removed in [39f5a8e](https://github.com/DiamondLightSource/httomo/pull/364/commits/39f5a8eb68e0fd611d0f2b9c5c3485f058261903)

yousefmoazzam: Sorry, to try and clarify, what I'm saying is that there's your additions that check `if not has_gpu` (starting at line 305):
```python
        if not has_gpu:
            section.max_slices = min(httomo.globals.MAX_CPU_SLICES, max_slices)
            return
```

but there's also something further down doing the exact same check and setting the max slices (starting at line 343):
```python
        if not has_gpu:
            section.max_slices = min(
                min(max_slices_methods), httomo.globals.MAX_CPU_SLICES
            )
```

What I'm asking is, why are there two checks for `if not has_gpu` to set `section.max_slices`?

dkazanc: yes, you're right. Shouldn't be there. 
[e1254db](https://github.com/DiamondLightSource/httomo/pull/364/commits/e1254db3eb82f94185ffb885f05c19e2ac736207)

PR #367, opened 2024-06-13
Add flag to specify number of frames per-chunk in intermediate hdf5 data

yousefmoazzam:
Built on top of work done in #271

This adds the `--frames-per-chunk` flag which enables the ability to:
- write the intermediate hdf5 data as chunked
- specify the number of frames (projections or sinograms) per-chunk

A vlaue of 0 will turn chunking off (ie, write the hdf5 data as contiguous), and values > 0 will be the number of frames per-chunk.

Acceptance criteria checklist:
- [x] Intermediate data is chunked as expected
- [x] Nicely handle the case when the frames per-chunk value provided exceeds the number of frames (h5py throws an error if this is the case, should be handled nicely in some way)
- [x] Add tests
- [x] All tests pass


yousefmoazzam:
To address the problem of if the value given to the `--frames-per-chunk` flag exceeds the length of the slicing dim (ie, the chunk shape exceeds the boundaries of the data to be written), when this happens, the frames per-chunk value used is 1 instead of the flag value.

If other behaviour is preferred, feel free to mention, I'll open this for review now.

httomo/cli.py
```diff
@@ -118,6 +118,12 @@ def check(yaml_config: Path, in_data_file: Path = None):
     default=514,
     help="Port on the host the syslog server is running on",
 )
+@click.option(
+    "--frames-per-chunk",
```
dkazanc, line 122: @ptim0626 @yousefmoazzam 
I wonder if we need more granular control over the chunk sizes here? For instance we would like to use a smaller chunk than the whole frame, e.g. `(1, 150, 150)`? Something like a tuple to pass instead of the number of frames. The first value in that tuple would be linked to the slicing dimension and the other two of non-slicing dimensions? 
Could that help with more versatile optimisation around the better write performance? And also the read speed in Dawn? 
Currently the run fails on Wilson if `frames-per-chunk > 1`

ptim0626: Considering performance, instead of decreasing the chunk size, I think increasing the chunk size would actually help.
(1) with compression, applying the filter to a single bigger chunk is usually quicker than multiple smaller chunks because of the overhead of threading for some filters llike Blosc
(2) larger chunks mean fewer metadata is required (the size of B-tree for looking up where the chunks are is smaller)

Choosing an optimal chunk size depends how we access the data, also a balance between read and write performance. I would say having one frame per chunk as the smallest unit with the option of increasing it via `--frames-per-chunk` is a good balance of performance and usability (does the general users know the implication of using a customised chunk shape?)

The recommended 1 MiB chunk size is about the default raw data chunk cache (rdcc) in hdf5, which is default to be 1 MiB. As long as we ensure we set the size of rdcc to be at least one chunk, it doesn't matter much for the chunk size, if it fits our access pattern. Even so, I don't think it impacts performance a lot, as most of the time (I guess) we just write/read the data once.

dkazanc: Thanks @ptim0626 , it makes sense. The general user won't be playing with `--frames-per-chunk` I reckon so it would be nice to fix a default size that fits our needs best.

PR #393, opened 2024-07-17
Fixes for memory estimations in the FBP algorithm

team-gpu:
(please adjust target branch - unsure if this should go to `dev` instead)

This fixes the memory estimation for the FBP algorithm. Here is a step by step breakdown of what happens in FBP in terms of allocation:

1. a 1D R2C FFT plan is created and used for [this real-to-complex FFT](https://github.com/dkazanc/ToMoBAR/blob/master/tomobar/fourier.py#L50) -> [fftplan_size](https://github.com/team-gpu/httomo/blob/bugfix/fbp-memory-estimation/httomo/methods_database/packages/external/httomolibgpu/supporting_funcs/recon/algorithm.py#L79)
2. This creates a [complex output array with half the width](https://github.com/team-gpu/httomo/blob/bugfix/fbp-memory-estimation/httomo/methods_database/packages/external/httomolibgpu/supporting_funcs/recon/algorithm.py#L89)
3. [An array for the filter is created](https://github.com/dkazanc/ToMoBAR/blob/master/tomobar/fourier.py#L53) (fixed, not dependent on number of slices) 
4. an [1D C2R FFT plan is created](https://github.com/team-gpu/httomo/blob/bugfix/fbp-memory-estimation/httomo/methods_database/packages/external/httomolibgpu/supporting_funcs/recon/algorithm.py#L97) and used for the reverse transform, creating [a real output array as well](https://github.com/team-gpu/httomo/blob/bugfix/fbp-memory-estimation/httomo/methods_database/packages/external/httomolibgpu/supporting_funcs/recon/algorithm.py#L107).

Note that at this point, we free all memory for the FFT plans and temporary data. So in the end, the max memory used will be the max of this first part and the second part, which [is reflected here](https://github.com/team-gpu/httomo/blob/bugfix/fbp-memory-estimation/httomo/methods_database/packages/external/httomolibgpu/supporting_funcs/recon/algorithm.py#L139). 

5. we have [a swapaxis call](https://github.com/dkazanc/ToMoBAR/blob/master/tomobar/methodsDIR_CuPy.py#L135) before giving the inputs to astra, which causes another allocation.
6. Before calling astra, we [allocate space for the reconstruction output](https://github.com/dkazanc/ToMoBAR/blob/master/tomobar/astra_wrappers/astra_base.py#L524), which is captured in the estimator [here](https://github.com/team-gpu/httomo/blob/bugfix/fbp-memory-estimation/httomo/methods_database/packages/external/httomolibgpu/supporting_funcs/recon/algorithm.py#L120).
7. then we call astra, which makes a copy of the input into a `cudaArray` for texture-based access. In the tests I have captured this [using mocks with side effects](https://github.com/team-gpu/httomo/blob/bugfix/fbp-memory-estimation/tests/test_backends/test_httomolibgpu.py#L434-L443), since these allocations aren't done by cupy and would be missed. 
8. The [check_kwargs](https://github.com/dkazanc/ToMoBAR/blob/master/tomobar/supp/suppTools.py#L399) seems to make a circular mask of the data, which does lots of small allocations / frees. The biggest one is the mask of `recon_size` elements with `int64` type. I'm putting this as a fixed cost, and [using `max` ](https://github.com/team-gpu/httomo/blob/bugfix/fbp-memory-estimation/httomo/methods_database/packages/external/httomolibgpu/supporting_funcs/recon/algorithm.py#L129-L131) as the filter from the first part will be gone and that memory is re-used.  

All the tests pass well now, without the strange multiplier. But please confirm with real world data if that is fully working.

### Further Notes / Future work

- the circular mask application depends on parameters and the mask radius should probably be considered in the estimators. But it's a comparably small fixed cost, so likely we can get away with it. 
- to reduce memory, we can destroy the FFT plan before calling the IFFT. That would reduce the memory of the filtersync a lot.
- wonder if we absolutely need the swapaxes calls - some experimentation here could help. That is since the copy to the cudaArray in astra makes another copy for example.
- the circular mask code looks very inefficient and can surely be optimised (and made less memory hungry). But I think both for memory and performance, it's not a bottleneck.

### DIFF
--- a/httomo/methods_database/packages/external/httomolibgpu/supporting_funcs/recon/algorithm.py
+++ b/httomo/methods_database/packages/external/httomolibgpu/supporting_funcs/recon/algorithm.py
@@ -110,20 +110,20 @@ def _calc_memory_bytes_FBP(
     # separate
     filtersync_size = input_slice_size + fftplan_slice_size + proj_f_slice + ifftplan_slice_size
     
-    # 6. we swap the axes before passing data to Astra
-    # https://github.com/dkazanc/ToMoBAR/blob/master/tomobar/methodsDIR_CuPy.py#L135
-    astra_input_swapaxis_slice = np.prod(non_slice_dims_shape) * np.float32().itemsize
+    # 6. we swap the axes before passing data to Astra in ToMoBAR
+    # https://github.com/dkazanc/ToMoBAR/blob/54137829b6326406e09f6ef9c95eb35c213838a7/tomobar/methodsDIR_CuPy.py#L135
+    pre_astra_input_swapaxis_slice = np.prod(non_slice_dims_shape) * np.float32().itemsize
     
     # 7. astra backprojection will generate an output array 
-    # https://github.com/dkazanc/ToMoBAR/blob/master/tomobar/astra_wrappers/astra_base.py#L524
+    # https://github.com/dkazanc/ToMoBAR/blob/54137829b6326406e09f6ef9c95eb35c213838a7/tomobar/astra_wrappers/astra_base.py#L524
     output_dims = _calc_output_dim_FBP(non_slice_dims_shape, **kwargs)
     recon_output_size = np.prod(output_dims) * np.float32().itemsize
     
     # 7. astra backprojection makes a copy of the input
     astra_input_slice_size = np.prod(non_slice_dims_shape) * np.float32().itemsize
     
-    ## now we calculate back projection memory
-    projection_mem_size = astra_input_swapaxis_slice + astra_input_slice_size + recon_output_size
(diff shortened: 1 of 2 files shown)


team-gpu:
Any idea about the check failures? Seems unrelated.

yousefmoazzam:
Eh, I wouldn't worry about the failing CPU tests, that's been failing for all PRs for a while now.

The PR from the `dev` branch has started to sort out the CI, which is the reason for the required status checks which are waiting here (they're waiting on workflows which are only available in the `dev` branch, in this branch the necessary workflow files aren't there), but for now we can probably ignore it.

Issue #383, opened 2024-07-08
Replace deprecated `ansi` param usage with `colors` param in logging


yousefmoazzam:
As-per the advice given when running tests:
```
============================================ warnings summary============================================
tests/method_wrappers/test_rotation.py: 16 warnings
  /httomo/httomo/utils.py:60: DeprecationWarning: The 'ansi' parameter is deprecated, please use 'colors' instead
    logger.opt(ansi=True).info("<yellow>{}</yellow>".format(output))

tests/method_wrappers/test_rotation.py::test_rotation_180
tests/method_wrappers/test_rotation.py::test_rotation_360
  /httomo/httomo/method_wrappers/rotation.py:191: RuntimeWarning: Mean of empty slice.
    flats1d = 1.0 if flats is None else flats.mean(0, dtype=np.float32)

tests/method_wrappers/test_rotation.py::test_rotation_180
tests/method_wrappers/test_rotation.py::test_rotation_360
  /opt/conda/lib/python3.10/site-packages/numpy/core/_methods.py:182: RuntimeWarning: invalid value encountered in divide
    ret = um.true_divide(

tests/method_wrappers/test_rotation.py::test_rotation_180
tests/method_wrappers/test_rotation.py::test_rotation_360
  /httomo/httomo/method_wrappers/rotation.py:192: RuntimeWarning: Mean of empty slice.
    darks1d = 0.0 if darks is None else darks.mean(0, dtype=np.float32)

tests/runner/test_task_runner.py: 16 warnings
  /httomo/httomo/utils.py:58: DeprecationWarning: The 'ansi' parameter is deprecated, please use 'colors' instead
    logger.opt(ansi=True).info("<green>{}</green>".format(output))

tests/runner/test_task_runner.py: 16 warnings
  /httomo/httomo/utils.py:62: DeprecationWarning: The 'ansi' parameter is deprecated, please use 'colors' instead
    logger.opt(ansi=True).info("<magenta>{}</magenta>".format(output))

tests/runner/test_task_runner.py::test_does_reslice_when_needed_and_reports_time
  /httomo/httomo/utils.py:56: DeprecationWarning: The 'ansi' parameter is deprecated, please use 'colors' instead
    logger.opt(ansi=True).info("<cyan>{}</cyan>".format(output))

tests/runner/test_task_runner.py::test_does_reslice_when_needed_and_reports_time
[...]

PR #396, opened 2024-07-18
Padding methods support in DB

team-gpu:
This adds support in the methods database to flag which are requiring padding. It also adds a new supporting function to calculate the padding values depending on the parameters, returning a tuple for the left and right pad values for each method.

The methods added so far are all the methods with the `3d` suffix from tomopy, nothing else yet. Please check if the padding calculation functions are actually correct for these, and check if some 3d methods might have been missed. But also note that it's easy to add more in future.

(It also fixes some pytest warnings that were flagged on my box)

### DIFF
--- a/tests/test_method_query.py
+++ b/tests/test_method_query.py
@@ -170,18 +170,21 @@ def test_database_query_calculate_output_dims(mocker: MockerFixture):
 
 
 def test_database_query_calculate_padding(mocker: MockerFixture):
+    SIZE_PARAMETER = 5
+    PADDING_RETURNED = (5, 5)
+
     class FakeModule:
         def _calc_padding_testmethod(size):
-            assert size == 5
-            return 5, 5
+            assert size == SIZE_PARAMETER
+            return PADDING_RETURNED
 
     importmock = mocker.patch("importlib.import_module", return_value=FakeModule)
     query = MethodsDatabaseQuery("sample.module.path", "testmethod")
 
-    pads = query.calculate_padding(size=5)
+    pads = query.calculate_padding(size=SIZE_PARAMETER)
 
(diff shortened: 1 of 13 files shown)


yousefmoazzam:
Fixes #383

tests/test_method_query.py
```diff
@@ -129,3 +167,21 @@ def _calc_output_dim_testmethod(non_slice_dims_shape, testparam):
...
+    importmock = mocker.patch("importlib.import_module", return_value=FakeModule)
+    query = MethodsDatabaseQuery("sample.module.path", "testmethod")
+
+    pads = query.calculate_padding(size=5)
+
+    importmock.assert_called_once_with(
+        "httomo.methods_database.packages.external.sample.supporting_funcs.module.path"
+    )
+
+    assert pads == (5, 5)
```
yousefmoazzam, line 187: Nitpick, take it or leave it: whenever I see a repeated constant I prefer to make one variable to reference everywhere, makes it less likely to get stuff wrong if needing to edit the test later.

Could define `LEFT_PADDING` and `RIGHT_PADDING` via something like `LEFT_PADDING = RIGHT_PADDING = 5`, and then use them where appropriate, would make the test slightly more understandable than seeing magic 5's in places?

Maybe also defining `SIZE`, so then the `assert size == 5` and the `pads = query.calculate_padding(size=5)` are more obviously connected to some `SIZE` variable?

yousefmoazzam (approved):
Looks good, and tests all pass for me locally. Only a few very minor things, take them or leave them :)

team-gpu:
I've addressed your comments (makes sense). Please go ahead and merge when you're happy with it otherwise.

PR #397, opened 2024-07-19
Removes old loader code and tests, but leaves a shell

team-gpu:
This removes a lot of dead code around the older loader, along with the tests for it. All tests pass for me, and full pipelines work too.


yousefmoazzam (approved):
Yep, can confirm that all tests are passing still, as well as the YAML template for the loader still being generate-able from the "shell" loader function

PR #399, opened 2024-07-22
Adds padding capability to blocks

team-gpu:
A small PR which simply adds interfaces and capabilities to the `DataSetBlock` for padding. It works as follows:

- a `padding` parameter is added to the constructor, which is a tuple of 2 `int` values. It means the number of padded slices to the left and to the right. 
- `block_shape` is adapted to include the padding, i.e. a `(2, 3)` padded block with a core shape of `(10, 3, 4)` and slicing dim `0` means it will  have `block_shape=(15, 3, 4)`.  This includes the 5 padding slices.
- `chunk_shape` is adapted in the same way - the shape is the previous chunk shape + the padding
- `chunk_index` and `global_index` are the indices within the chunk data or global data, respectively, where the block starts, so this includes the padding as well. This means there can be negative indices as well - i.e. a block or chunk with `padding=(2, 2)` can have a starting index of `-2` and can run "over the boundary" by 2. This makes sure that the `data` field of the block plugs into the right position of the global data in all cases.
- However, the `global_shape` has not been adapted for padding - I believe that should always be the true global data, regardless of padding. The padding feature is something that's required for chunking and blocking the global data for consistency, but has nothing to do with the full global shape. 

In addition, `*_unpadded` properties have been added to the blocks for convenience, so that the indexes and shapes of the core area of this block can be easily retrieved. This is to help with writing to the dataset store, etc.

The tests have been adapted to handle padding, consistency checks modified, and all existing tests are still working (as padding defaults to `(0, 0)`.

### DIFF
--- a/httomo/runner/dataset.py
+++ b/httomo/runner/dataset.py
@@ -25,10 +25,10 @@ class DataSetBlock(BaseBlock, BlockIndexing):
         padding: Tuple[int, int] = (0, 0),
     ):
         """Constructs a data block for processing in the pipeline in high troughput runs.
-        
+
         Parameters
         ----------
-        
+
         data: ndarray
             A numpy or cupy array, 3D, holding the data represented by this block.
         aux_data: AuxiliaryData
@@ -44,7 +44,7 @@ class DataSetBlock(BaseBlock, BlockIndexing):
             A chunk is a unit of the global data that is handled by a single MPI process, while a block
             might be a smaller part than the chunk.
         global_shape:  Optional[Tuple[int, int, int]]
-            The shape of the global data across all processes. If not given, it assumes this block 
+            The shape of the global data across all processes. If not given, it assumes this block
             represents the full global data (no slicing done).
         chunk_shape: Optional[Tuple[int, int, int]]
             The shape of the chunk that this block belongs to. If not given, it assumes this block
@@ -52,7 +52,7 @@ class DataSetBlock(BaseBlock, BlockIndexing):
         padding: Tuple[int, int]
             Padding information - holds the number of padded slices before and after the core area of the block,
             in slicing dimension. If not given, no padding is assumed.
-            
+
             Note that the padding information should be added to the data's shape, i.e. block_start, chunk_start,
             chunk_shape, and the data's shape includes the padded slices. And therefore block_start or chunk_start
             may have negative values of up to -padding[0]. The global_shape is not adapted for padding.
@@ -103,9 +103,16 @@ class DataSetBlock(BaseBlock, BlockIndexing):
             > self.global_shape[self.slicing_dim]
         ):
             raise ValueError("chunk spans beyond the global data boundaries")
-        if any(self.chunk_shape[i] > self.global_shape[i] for i in range(3) if i != self.slicing_dim):
+        if any(
+            self.chunk_shape[i] > self.global_shape[i]
+            for i in range(3)
+            if i != self.slicing_dim
+        ):
             raise ValueError("chunk shape is larger than the global shape")
-        if self.chunk_shape[self.slicing_dim] - self.padding[0] - self.padding[1] > self.global_shape[self.slicing_dim]:
+        if (
+            self.chunk_shape[self.slicing_dim] - self.padding[0] - self.padding[1]
+            > self.global_shape[self.slicing_dim]
+        ):
             raise ValueError("chunk shape is larger than the global shape")
         if any(self.shape[i] > self.chunk_shape[i] for i in range(3)):
             raise ValueError("block shape is larger than the chunk shape")
@@ -187,3 +194,37 @@ class DataSetBlock(BaseBlock, BlockIndexing):
     @property
     def padding(self) -> Tuple[int, int]:
         return self._padding
+
+    @property
+    def shape_unpadded(self) -> Tuple[int, int, int]:
+        return self._correct_shape_for_padding(self.shape)
+
+    @property
+    def chunk_index_unpadded(self) -> Tuple[int, int, int]:
+        return self._correct_index_for_padding(self.chunk_index)
+
+    @property
+    def chunk_shape_unpadded(self) -> Tuple[int, int, int]:
+        return self._correct_shape_for_padding(self.chunk_shape)
+
+    @property
+    def global_index_unpadded(self) -> Tuple[int, int, int]:
+        return self._correct_index_for_padding(self.global_index)
+
+    def _correct_shape_for_padding(
+        self, shape: Tuple[int, int, int]
+    ) -> Tuple[int, int, int]:
+        if not self.padding:
+            return shape
+        shp = list(shape)
+        shp[self.slicing_dim] -= self.padding[0] + self.padding[1]
+        return make_3d_shape_from_shape(shp)
+
+    def _correct_index_for_padding(
+        self, index: Tuple[int, int, int]
+    ) -> Tuple[int, int, int]:
+        if not self.padding:
+            return index
+        idx = list(index)
+        idx[self.slicing_dim] += self.padding[0]
+        return make_3d_shape_from_shape(idx)
--- a/httomo/runner/dataset.py
+++ b/httomo/runner/dataset.py
@@ -85,6 +85,8 @@ class DataSetBlock(BaseBlock, BlockIndexing):
         self._check_inconsistencies()
 
     def _check_inconsistencies(self):
+        if self.padding[0] < 0 or self.padding[1] < 0:
+            raise ValueError("padding values cannot be negative")
         if self.chunk_index[self.slicing_dim] + self._padding[0] < 0:
             raise ValueError("block start index must be >= 0")
         if (
--- a/httomo/runner/dataset.py
+++ b/httomo/runner/dataset.py
@@ -110,12 +110,16 @@ class DataSetBlock(BaseBlock, BlockIndexing):
             for i in range(3)
             if i != self.slicing_dim
         ):
-            raise ValueError("chunk shape is larger than the global shape")
+            raise ValueError(
+                "chunk shape is larger than the global shape in non-slicing dimensions"
+            )
         if (
             self.chunk_shape[self.slicing_dim] - self.padding[0] - self.padding[1]
             > self.global_shape[self.slicing_dim]
         ):
-            raise ValueError("chunk shape is larger than the global shape")
+            raise ValueError(
+                "chunk shape is larger than the global shape in slicing dimension"
+            )
         if any(self.shape[i] > self.chunk_shape[i] for i in range(3)):
             raise ValueError("block shape is larger than the chunk shape")
         if any(
--- a/httomo/runner/dataset.py
+++ b/httomo/runner/dataset.py
@@ -24,7 +24,7 @@ class DataSetBlock(BaseBlock, BlockIndexing):
         chunk_shape: Optional[Tuple[int, int, int]] = None,
         padding: Tuple[int, int] = (0, 0),
     ):
-        """Constructs a data block for processing in the pipeline in high troughput runs.
+        """Constructs a data block for processing in the pipeline in high throughput runs.
 
         Parameters
         ----------
--- a/httomo/runner/dataset.py
+++ b/httomo/runner/dataset.py
@@ -21,17 +21,21 @@ class DataSetBlock(BaseBlock):
         chunk_start: int = 0,
         global_shape: Optional[Tuple[int, int, int]] = None,
         chunk_shape: Optional[Tuple[int, int, int]] = None,
+        padding: Tuple[int, int] = (0, 0),
     ):
         super().__init__(data, aux_data)
         self._slicing_dim = slicing_dim
         self._block_start = block_start
         self._chunk_start = chunk_start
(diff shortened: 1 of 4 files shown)


httomo/runner/dataset.py
```diff
@@ -41,28 +79,50 @@ def __init__(
...
+            self.global_index[self.slicing_dim]
+            + self.shape[self.slicing_dim]
+            - self._padding[1]
+            > self.global_shape[self.slicing_dim]
+        ):
             raise ValueError("chunk spans beyond the global data boundaries")
-        if any(self.chunk_shape[i] > self.global_shape[i] for i in range(3)):    
+        if any(self.chunk_shape[i] > self.global_shape[i] for i in range(3) if i != self.slicing_dim):
+            raise ValueError("chunk shape is larger than the global shape")
+        if self.chunk_shape[self.slicing_dim] - self.padding[0] - self.padding[1] > self.global_shape[self.slicing_dim]:
```
yousefmoazzam, line 108: After some thinking, I _think_ I understand why the second `if` condition has been added here (to catch an _unpadded_ chunk that is larger than the global shape in the slicing dim?).

However, both checks produce the exact same error message of "chunk shape is larger than the global shape": even if it's technically a correct error message for both cases, I'd prefer there to be some distinction between the two message, so then someone reading the code would have a better chance of understanding the difference between the two checks.

Are the two checks something like:
- first check: "is the chunk (for both padded and unpadded cases) larger than the global data in the non-slicing dims?"
- second check: "is the chunk (for both padded and unpadded) larger than the global data in the slicing dim?"

(Maybe not quite, I'm not sure)

tests/runner/test_dataset_block.py
```diff
@@ -103,6 +127,113 @@ def test_partial_block_for_chunked_data(slicing_dim: Literal[0, 1, 2], last_in_c
...
+        slicing_dim=slicing_dim,
+        block_start=start_index,
+        chunk_start=10,
+        global_shape=global_shape,
+        chunk_shape=chunk_shape,
+        padding=(2, 2),
+    )
+
+    assert block.is_padded is True
+    assert block.padding == (2, 2)
```
yousefmoazzam, line 159: Sorry for the nitpick: could you add a `padding = (2, 2)` var to be reused, just to make things a tad more explanatory, like in `test_partial_block_with_padding_global_boundaries`?

tests/runner/test_dataset_block.py
```diff
@@ -103,6 +127,113 @@ def test_partial_block_for_chunked_data(slicing_dim: Literal[0, 1, 2], last_in_c
...
```
yousefmoazzam, line 192: Also here could `padding = (2, 2)` be added

tests/runner/test_dataset_block.py
```diff
@@ -103,6 +127,113 @@ def test_partial_block_for_chunked_data(slicing_dim: Literal[0, 1, 2], last_in_c
...
+        chunk_start=chunk_start,
+        global_shape=global_shape,
+        chunk_shape=chunk_shape,
+        padding=padding,
+    )
+
+    assert block.is_padded is True
+    assert block.padding == padding
+    assert block.global_index == tuple(global_index)
+    assert block.is_last_in_chunk is (boundary == "after")
```
yousefmoazzam, line 234: I see some magic numbers, like 6, 30, and 20, that get repeated in several places: could they be put into vars with names that explain what they are, and then reused where necessary?

tests/runner/test_dataset_block.py
```diff
@@ -103,6 +127,113 @@ def test_partial_block_for_chunked_data(slicing_dim: Literal[0, 1, 2], last_in_c
...
+@pytest.mark.parametrize("slicing_dim", [0, 1, 2])
+def test_partial_block_for_chunked_data_with_padding_center(
+    slicing_dim: Literal[0, 1, 2]
+):
+    block_shape = [10, 10, 10]
+    block_shape[slicing_dim] = 6
+    start_index = 3
+    data = np.ones(block_shape, dtype=np.float32)
+    global_index = [0, 0, 0]
+    global_index[slicing_dim] = 10 + start_index
```
yousefmoazzam, line 139: Is this 10 connected to the 10 on line 152 where there is `chunk_start=10`? (If it is, would be good to put it in a var for reuse+understanding, but if not then no worries)

tests/runner/test_dataset_block.py
```diff
@@ -103,6 +127,113 @@ def test_partial_block_for_chunked_data(slicing_dim: Literal[0, 1, 2], last_in_c
...
+@pytest.mark.parametrize("boundary", ["before", "after"])
+def test_partial_block_for_chunked_data_with_padding_chunk_boundaries(
+    slicing_dim: Literal[0, 1, 2], boundary: Literal["before", "after"]
+):
+    block_shape = [10, 10, 10]
+    block_shape[slicing_dim] = 6
+    start_index = -2 if boundary == "before" else 6
+    data = np.ones(block_shape, dtype=np.float32)
+    global_index = [0, 0, 0]
+    global_index[slicing_dim] = 10 + start_index
```
yousefmoazzam, line 172: Same question here about the 10

yousefmoazzam (commented):
In general looks good, thanks! Lots of little questions and some nitpicks, apologies... :sweat_smile:

httomo/runner/dataset.py
```diff
@@ -41,28 +45,50 @@ def __init__(
...
-        global_index[slicing_dim] += chunk_start + block_start
+        global_index[slicing_dim] += chunk_start + block_start + padding[0]
         self._global_index = make_3d_shape_from_shape(global_index)
 
         self._check_inconsistencies()
-        
+
     def _check_inconsistencies(self):
-        if self.chunk_index[self.slicing_dim] < 0:
+        if self.chunk_index[self.slicing_dim] + self._padding[0] < 0:
```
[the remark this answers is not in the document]
team-gpu: done in 6ae95fc602

tests/runner/test_dataset_block.py
```diff
@@ -37,6 +37,28 @@ def test_full_block_for_global_data():
     assert block.flat.dtype == data.dtype
 
 
+def test_full_block_for_global_data_with_padding():
+    padding = (2, 2)
+    data = np.ones((14, 10, 10), dtype=np.float32)
+    angles = np.linspace(0, math.pi, 10, dtype=np.float32)
+    block = DataSetBlock(data=data, aux_data=AuxiliaryData(angles=angles), padding=(2, 2), slicing_dim=0,
+                         block_start=-2, chunk_start=-2)
```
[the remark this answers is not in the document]
team-gpu: all fixed in fddde0d05

httomo/runner/dataset.py
team-gpu: fixed in fddde0d05

tests/runner/test_dataset_block.py
team-gpu: I reworked all the magic numbers in the tests - please see if it's easier to follow now.

tests/runner/test_dataset_block.py
team-gpu: see above

tests/runner/test_dataset_block.py
```diff
@@ -131,107 +169,167 @@ def test_partial_block_for_chunked_data(
...
+    padded_chunk_shape = make_3d_shape_from_shape(padded_chunk_shape_t)
+
+    unpadded_block_slices = 2
+    unpadded_block_shape_t = list(unpadded_chunk_shape)
+    unpadded_block_shape_t[slicing_dim] = unpadded_block_slices
+    unpadded_block_shape = make_3d_shape_from_shape(unpadded_block_shape_t)
+    padded_block_shape_t = unpadded_block_shape_t
+    padded_block_shape_t[slicing_dim] += padding[0] + padding[1]
+    padded_block_shape = make_3d_shape_from_shape(padded_block_shape_t)
+    block_start_index = 3
```
yousefmoazzam, line 190: Firstly, is the code comment I've suggested correct?

```suggestion
    block_start_index = 3  # start of block within chunk, without padding
```

If so, this is just for the sake of clarifying things, because I got confused initially why `block_start_index` was being shifted by the "before" padding value later in the test.

tests/runner/test_dataset_block.py
```diff
@@ -131,107 +169,167 @@ def test_partial_block_for_chunked_data(
...
+    assert block.chunk_shape[slicing_dim] == padded_chunk_shape[slicing_dim]
+    assert block.chunk_shape_unpadded[slicing_dim] == unpadded_chunk_shape[slicing_dim]
+    assert (
+        block.global_index[slicing_dim]
+        == chunk_start_index + padding[0] + block_start_index
+    )
+    assert (
+        block.global_index_unpadded[slicing_dim]
+        == chunk_start_index + padding[0] + block_start_index + padding[0]
+    )
```
yousefmoazzam, line 220: I'm trying to understand what `block.global_index` is, now that padding has been introduced.

From the first assertion here, given that it includes one shift of `padding[0]`, naively it seems _something_ like "`block.global_index` contains padding for the chunk that the block belongs to, but not padding for the block", is that right?

And if I'm honest, I'm not 100% sure if my question makes sense, I think I may also have some confusion about padding applied to a block as far as its consequences for index shifting. Stuff like "for the global index, if you pad a block, do you need to account for the index-shift in both `block_start_index` and `chunk_start_index`" come to mind...

Any clarifying comments would be greatly appreciated :sweat_smile: (And sorry for the dumb questions!)

yousefmoazzam (commented):
Almost there I think, the revamping of the test to explain the magic numbers is more readable now!

Two of the three comments I've made are me being confused about how index-shifting is working (for `chunk_index` and `global_index`), rather anything code-specific. Any pointers to better my understanding would be great  :slightly_smiling_face:

tests/runner/test_dataset_block.py
```diff
@@ -131,107 +169,167 @@ def test_partial_block_for_chunked_data(
...
         slicing_dim=slicing_dim,
-        block_start=start_index,
-        chunk_start=10,
+        block_start=block_start_index,
+        chunk_start=chunk_start_index,
         global_shape=global_shape,
-        chunk_shape=chunk_shape,
-        padding=(2, 2),
+        chunk_shape=padded_chunk_shape,
+        padding=padding,
```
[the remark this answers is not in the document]
team-gpu: It could indeed deduce the padded chunk shape from the padding parameter. But then I would find it confusing when asking the block for its chunk_shape, as that would return something else then what's passed to the constructor... The way it is now is:

```python
block = DataSetBlock(..., chunk_shape=XXX, padding=..., ...)
assert block.chunk_shape == XXX
```

If we deduce the chunk shape with padding included from that parameters, it would look like this (pseudo-code):

```python
block = DataSetBlock(..., chunk_shape=XXX, padding=pad ...)
assert block.chunk_shape == XXX + pad
```

To me, the first part is more consistent.

The alternative of deducing padding from the chunk shape given and the block's shape won't work either, as we don't know how much of it is before and how much is after.

tests/runner/test_dataset_block.py
team-gpu: The comment is not correct here - it's including the padding. It's the start index of the block's data, incl. padding. That's reflected in the later assert: 

```python
assert block.chunk_index[slicing_dim] == block_start_index
```

since the block's properties for `chunk_index` are including the padded slices. It's the index of the data in the block within the chunk, and the data includes the padded slices.

tests/runner/test_dataset_block.py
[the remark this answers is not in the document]
yousefmoazzam: In the constructor there's logic to handle when `global_shape` is passed in and padding needs to be accounted for:
```python
        if global_shape is None:
            global_shape_t = list(data.shape)
            global_shape_t[slicing_dim] -= padding[0] + padding[1]
            self._global_shape = make_3d_shape_from_shape(global_shape_t)
        else:
            self._global_shape = global_shape
```

Couldn't a similar thing be done using the `chunk_shape` and `padding` parameters of the constructor to set `self._chunk_shape` accordingly, so then the `self.chunk_shape` property still returns the same as it did before, just that the constructor sets it (rather than the creator of the block)?

tests/runner/test_dataset_block.py
team-gpu: The global index is the index of the block's data within the global data. The block's data includes padding. But I agree the test is confusing - because `chunk_start_index` also includes padding. That's why padding has to added in this assert, otherwise it's applied twice. This is the relation:

```python
block_start_index = block_core_start_index - padding[0]
chunk_start_index = chunk_core_start_index - padding[0]
block_global_index = chunk_core_start_index + block_core_start_index - padding[0]
```

substituting, we get:

```python
block_global_index = chunk_core_start_index + bock_start_index
                   = chunk_start_index + padding[0] + block_start_index
```

Let me change variables in the test a bit to make this more clear....

tests/runner/test_dataset_block.py
yousefmoazzam: Gotcha, I wasn't at all confident about the comment I suggested, thanks for clarifying.

yousefmoazzam (approved):

PR #401, opened 2024-07-24
DataSetStore with padding support

team-gpu:
This PR adds support for padded data to the `DataSetStoreWriter` and `DataSetStoreReader` classes. In particular:

- writing blocks that include padding to the `DataSetStoreWriter` results on the padding slices to be dropped and only the core area to be written. 
- the `DataSetStoreWriter.make_reader` method gained a parameter for padding. The idea is that the task runner will create the reader this way for the next section, and it knows if that section needs padding and how much. 
- the `DataSetStoreReader` correctly handles padding, i.e.:
   - the reader properties `chunk_shape` and `global_index` are adapted to include the padding
   - when `read_block(start, length)` is called, it automatically generates a padded block. The parameters `start` and `length` are regarding the core area, but the shape of the returned block is `length + padding[0] + padding[1]`.
   - if the global data is stored in a file, the padding area is read directly from it, and the edges are extrapolated using the `_extrapolate_before` and `_extrapolate_after` methods
   - if the chunk's data is stored in memory, we need to ask the neighbour MPI processes for the padding area. This has been solved by actually including this area in `self._data` at construction time of the reader. The MPI exchange is done when the reader is constructed, so that the chunk + padding slices is available in `self._data`. This includes also the extrapolation at the boundaries. 

A set of tests confirms correct behaviour:
- file-based with padding: testing padding and extrapolation in center, front, back slices, and also if there is only one process (both edges are extrapolated)
- RAM-based with padding: testing MPI exchanges of neighbourhoods + boundary extrapolation
- both of the above with reslice, to make sure that the parameters and shapes are working after reslicing too.

### DIFF
--- a/httomo/data/dataset_store.py
+++ b/httomo/data/dataset_store.py
@@ -231,7 +231,11 @@ class DataSetStoreWriter(ReadableDataSetSink):
         sendBuffer = np.zeros(1, dtype=bool)
         recvBuffer = np.zeros(1, dtype=bool)
         try:
-            self._data = self._create_numpy_data(self.chunk_shape, block.data.dtype)
+            self._data = self._create_numpy_data(
+                unpadded_chunk_shape=block.chunk_shape_unpadded,
+                padded_chunk_shape=block.chunk_shape,
+                dtype=block.data.dtype,
+            )
         except MemoryError:
             sendBuffer[0] = True
 
@@ -253,17 +257,21 @@ class DataSetStoreWriter(ReadableDataSetSink):
             )
 
     def _create_numpy_data(
-        self, chunk_shape: Tuple[int, int, int], dtype: DTypeLike
+        self,
+        unpadded_chunk_shape: Tuple[int, int, int],
+        padded_chunk_shape: Tuple[int, int, int],
+        dtype: DTypeLike,
     ) -> np.ndarray:
         """Convenience method to enable mocking easily"""
+        unpadded_chunk_bytes = np.prod(unpadded_chunk_shape) * np.dtype(dtype).itemsize
+        padded_chunk_bytes = np.prod(padded_chunk_shape) * np.dtype(dtype).itemsize
         if (
             self._memory_limit_bytes > 0
-            and np.prod(chunk_shape) * np.dtype(dtype).itemsize
-            >= self._memory_limit_bytes
+            and unpadded_chunk_bytes + padded_chunk_bytes >= self._memory_limit_bytes
         ):
             raise MemoryError("Memory limit reached")
 
-        return np.empty(chunk_shape, dtype)
+        return np.empty(unpadded_chunk_shape, dtype)
 
     def _create_h5_data(
         self,
--- a/httomo/data/dataset_store.py
+++ b/httomo/data/dataset_store.py
@@ -12,6 +12,7 @@ from httomo.runner.dataset_store_interfaces import (
     ReadableDataSetSink,
 )
 from mpi4py import MPI
+from mpi4py.util import dtlib
 import numpy as np
 from numpy.typing import DTypeLike
 import weakref
@@ -358,6 +359,9 @@ class DataSetStoreReader(DataSetSource):
             chunk_shape_t[self.slicing_dim] += self._padding[0] + self._padding[1]
             self._chunk_shape = make_3d_shape_from_shape(chunk_shape_t)
 
+            if not source.is_file_based:
+                self._exchange_neighbourhoods()
+
         source.finalize()
 
     @property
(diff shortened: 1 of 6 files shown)


httomo/data/dataset_store.py
```diff
@@ -411,32 +436,211 @@ def _reslice(
...
+            )
+
+    def _extend_data_for_padding(self, core_data: np.ndarray) -> np.ndarray:
+        padded_data = np.empty(self._chunk_shape, self._data.dtype)
+        core_slices = [slice(None), slice(None), slice(None)]
+        core_slices[self._slicing_dim] = slice(
+            self._padding[0], self._chunk_shape[self._slicing_dim] - self._padding[1]
+        )
+        padded_data[core_slices[0], core_slices[1], core_slices[2]] = core_data
+        return padded_data
```
yousefmoazzam, line 589: Question: what happens if there's not enough RAM to hold both:
1. an array the size of the "core" of the chunk
2. an array the size of the padded version of the chunk

## The reasoning behind the question

Whether the store is file-based or RAM-based is determined at the time the _writer_ attempts to write the first block of the first section to the store. The max RAM required is determined by the writer by the size of a single chunk. This is before the reader object is created.

Suppose there's enough RAM to hold the chunk, so the writer determines that the store can be RAM-based when it attempts to write the first block to the store. When a reader object is created, the `_exchange_neighbourhoods()` method is called (since RAM-based requires each process to get the neighbourhoods from the chunks of the adjacent processes).

At that time, the reader will attempt to create another array with the padded chunk shape. At this point, both an array of the core chunk shape, and an array of the padded chunk shape, exist simultaneously. This means that the RAM requirement of the process has now doubled.

At the time of the writer deciding the store can be RAM-based, the max RAM needed was a single chunk. At the later time of the reader being created though, it seems like the RAM requirements are double what they were at the time the writer was deciding whether the store should be file-based or RAM-based.

When the `self._data` variable gets reassigned to the array representing the padded chunk, the array representing the core chunk should be eligible for the garbage collector to clean up. However, this reassignment doesn't occur until _after_ the `_extend_data_for_padding()` has returned the padded chunk array.

What happens if there's enough RAM to hold a single chunk, but not enough RAM to hold essentially 2 copies of the chunk? (This is in the short period of time where the `self._data` variable hasn't yet been reassigned and we are still in the execution of `_extend_data_for_padding()` after `padded_data` has been defined)

team-gpu: This will indeed be a problem and could happen... Thanks for pointing this out. To add to that, it's also possible that the writer of the previous section could fit a chunk into RAM, but chunk + padding slices wouldn't fit (even without the doubling of the memory issue that you pointed out). 

I think the chances are rather low, because:
- the reader of the previous section + the blocks used during process also had occupied memory, which became available at the time the next reader is constructed (**IMPORTANT**: TaskRunner should make sure that the previous reader is destructed _before_ the next one is created)
- the section which starts with this reader will also need memory to process blocks and store results. If it's that tight, it might blow up memory in the section that follows (can't construct blocks)

BUT: The chance is real, and it could happen in real pipelines - it needs to be handled. 

### Suggested Solution

- let the writer of the previous section know about the padding required for the chunks - and make sure it allocates enough space for the padded chunk directly. It does not need to write into the padded area - just make sure the memory is allocated. 
- if it fails to allocate the chunk with padding in RAM, it will go to a file and the issue is not there anyway
- the writer's `write_block` would need to offset the data by the padded slices. 
- the reader can then assume if it's RAM-based that `source._data` includes space for the padding and has the right data at the core already. It would just need to fill the boundaries as it already does. (Note that indexes of boundaries will also be affected in neighbouring processes.)

_Note_: This means the padding slices for the next section needs to be calculated as part of the previous section's setup, but that should be possible... unless the padding slices depend on a side output of the previous section (I think remote, but should be checked and perhaps flagged with a clear error, or an assert). 

Does that make sense?

yousefmoazzam: Thanks for the second opinion on this, always good to get other's thoughts on if something is an issue or not :slightly_smiling_face: 

I think I understand the suggested solution, and thank you for the steps that can be done in order to implement it. Is it ok if I describe how I think the suggested solution solves the issue of the "two copies in RAM", to ensure I understand it, and then you can correct me if/as necessary?

- when a writer object writes the first block, it will allocate an array to hold not the "core" of the chunk, but an array to hold a padded version of the chunk
  - caveat: the padding applied to the chunk needs to be the padding that the next section requires, so the padded chunk array created by a writer for section `n` needs to have a shape reflecting the padding required for section `n+1`
- then, when it comes to the reader of the next section, it no longer needs to create a second array which has the shape of the padded chunk
- therefore, we no longer have the "two copies of the chunk in RAM" issue

### Another question that popped in my mind from the suggested solution

I don't _think_ this is an issue, but it's worth mentioning to get a second opinion: if there's a reslice between section `n` and `n+1`, does this affect things? For example, section `n` is working on projections (`slicing_dim=0`), and section `n+1` is working on sinograms (`slicing_dim=1`) and needs padding.

The writer for section `n` should allocate an array whose shape reflected the padding in the slicing dim for the _next section_, rather than for the current section. Specifically, the writer for section `n` needs to apply padding to the chunk in `slicing_dim=1` (because section `n+1` has `slicing_dim=1`), even though section `n` has `slicing_dim=0`.

I don't think I see a problem, the blocks for section `n` can still be written to the "core part" of the padded chunk array without issue. All I see is that it's simply a bit odd when thinking about it at first, to "apply padding" during execution of section `n` in a dimension different to the slicing dim of section `n`.

Thoughts?

team-gpu: Your understanding of the solution is correct.

About reslicing - you're raising a good point there:
- if there is a reslice in the next section, there's actually no point to create a padded chunk: padding in this direction is useless if the next section requires padding in the other.
- reslice makes a copy of the chunk in any case for the send/receive buffers - the issue with 2x the memory is there also, and can't really be avoided...
- reslice can possibly be modified to reslice the data into the core area of an already-padded array in the new direction? This could create performance issues though, since this core area in the general case might not be contiguous in memory and MPI exchanges may not be efficient with that. 
- Or somehow be aware of the padding itself and exchange the neighbourhoods directly with the reslice operation (this is an optimisation and might be a bit complicated)?

Another option could be to completely ignore this whole issue also, risking out of memory errors on the CPU in some cases... ;-)

Or just make sure that the memory for the padded array that is needed by the next reader is pre-allocated before the first block is written to the sink - just to make sure there is enough room in RAM to hold it, handle reslices, etc. If that allocation fails, it should to go file in any case. But don't really use this array - that's purely to ensure it fits in memory. It can be temporarily dropped before calling the `reslice` function, to make room for the copies that reslice does.

yousefmoazzam:
For the sake of visibility, this is just a follow up on [this discussion](https://github.com/DiamondLightSource/httomo/pull/401#discussion_r1691617937):
- the chosen solution after some discussion offline was to have the writer take into account the padded chunk shape of the following section
- after that discussion, an additional small change has been made to the writer to provide a starting point for a solution to the above
  - the change to the writer does not fully implement the solution, it is using the padded chunk shape for the _current section_
  - a few more changes need to be done to be able to pass the padded chunk shape for the next section to the writer
  - when those changes are made, the means through which the writer gets the padded chunk shape will likely change, so for now, the writer has been left as-is
  - the necessary changes to the writer will be done alongside the changes when the writer is given the padded chunk shape info

PR #408, opened 2024-08-01
Implement parallelised parameter sweep runs

yousefmoazzam:
Main changes:
- distribute sweep values across MPI processes (ie, each process executes a subset of the full set of sweep values)
- modify image wrapper slightly to define the `offset` parameter only if it hasn't been set (see message of 527ca81 for more info)

Note: the intermediate data saver wrapper has _not_ been modified to be able to save hdf5 file output in parallel. Meaning, the use of the save intermediate wrapper currently is not tested in the context of a parallel sweep run.

Acceptance criteria
- [x] all tests pass
- [x] logging has been modified appropriately to reflect the fact that now a sweep run can be executed by multiple processes
- [x] images written to filesystem are verified to be correct


yousefmoazzam:
As a note on the logging, this is what it looks like currently in the terminal output:
```
(/dls/science/users/twi18192/conda-envs/httomo) [twi18192@cs05r-sc-gpu05-17 httomo (parallel-sweeps)]$ mpi run -n 4 python -m httomo run tests/test_data/tomo_standard.nxs sweep.yaml /dls/tmp/twi18192/sweep-tests/
See the full log file at: /dls/tmp/twi18192/sweep-tests/01-08-2024_12_02_30_output/user.log
Loading data with shape (180, 7, 160)
Running data_reducer (httomolib)
Running normalize (httomolibgpu)
Running remove_all_stripe (httomolibgpu)
Running FBP (httomolibgpu)
    Beginning parameter sweep
    Parameter name: center
    Total number of values across all processes: 10
    Values executed in this process: 2
      0%|          | 0/2 [00:00<?, ?value/s, center=100]
     50%|#####     | 1/2 [00:00<00:00,  1.48value/s, center=101]
    Finished parameter sweep
Running save_to_images (httomolib)
Pipeline finished. Took 1.745s
```

dkazanc:
thanks @yousefmoazzam. 
I've noticed that the `axis` parameter is set to one as opposed to `auto` (defined by the slicing dimension). When set to `auto`, however, it saves not what we want. For instance, for 10 values, it would save 2560 of `10 x 2560` images. I suspect it will be slightly tricky for users to operate with the correct slicing dimension for sweeps if the `axis` parameter is not automatic. 
```
- method: save_to_images
  module_path: httomolib.misc.images
  parameters:
    subfolder_name: images
    axis: 1
```

This leads me to the bigger question. _Do we want users to use `save_to_images` method explicitly in the pipeline in order to save the `sweep` images?_ 
Initially I thought that we could hide the method of saving images from the users. So if `sweep` is enabled in the method, `save_to_images` will be automatically inserted (similar to `data_reducer`) with the correct `axis` value as in the issue above. So the users would not need to worry of adding the image saving method into the pipeline. A bit like [write_center](https://tomopy.readthedocs.io/en/stable/api/tomopy.recon.rotation.html#tomopy.recon.rotation.write_center) method in TomoPy, where saving images is the part of the method itself. 
In that case, in the insertion method, we need to remove `save_to_images` if it  already present in the pipeline after sweep. And catching this in our yaml checker.

yousefmoazzam:
> This leads me to the bigger question. _Do we want users to use `save_to_images` method explicitly in the pipeline in order to save the `sweep` images?_

This is a pretty reasonable question: even for myself (ie, a developer), thinking a bit about the `axis` parameter value was annoying/tedious. Given that, I'd probably say that it's a good idea to try and take care of this for the user if possible!

Based on detecting if the pipeline is a sweep one or not, automatic insertion of the image wrapper into the `Pipeline` in the `TransformLayer` seems like a decent approach to try.

Truth be told, I was hoping to keep the `UiLayer` and `TransformLayer`unaware of if the pipeline was a sweep one or high throughput one, for the sake of making them as dumb as possible and not doing anything complicated. However, perhaps that is too much to ask when there's this sort of help that needs to be given to users to ensure correct pipeline configuration.

dkazanc:
thanks. yes, I think we need to simplify things for users unfortunately in favour of making things slightly nonlinear in some functions. I guess the good news is that the sweeping thing is a one-off hopefully and anything like this is not planned/needed as I'm aware. So it is not that bad to have some divergences from the main for sweep. 
I'll see myself what I can do here and ask for help if stuck. 
This and then the value masking. After this we can finalise this.

yousefmoazzam:
Sure, that's fair enough.

You know, I've been quite fond of the recent approach I've been taking (following on from Jorg :slightly_smiling_face: hehe) of having smaller PRs that are short-lived/get merged quickly.

Since the sweep running in parallel is working at its core (which is what the PR was intended for), and the next things to add are conveniences for the user, we could merge this now and cover the user-conveniences in a separate PR, what do you think?

It helps mitigate to some degree branches going stale and constantly needing to be updated from `main` over time, and it makes each PR more focused, which are both nice.

dkazanc:
sure that makes sense, we can merge this first.

yousefmoazzam:
Ok, cool! Let's see, the only problem I can see is that there are some failing FBP memory hook tests, are these expected/ok to be failing?

dkazanc:
yes, they are failing in `main` as well. We will get them sorted when we decide on how we're going to deal with the memory estimation for FBP.

PR #417, opened 2024-08-14
Ensure no Python objects pointed to CuPy array before cleaning its memory pool 

ptim0626:
This PR removes the reference to CuPy array inside `DataSetBlock` before cleaning up by the CuPy memory pool. This ensures no Python object refers to the underlying CuPy memory pointer, so the CuPy memory pool knows the memory pointer is not referenced by some Python objects, and the memory pool can be properly cleaned up.

A simple deleter of `data` attribute in `DataSetBlock` is implemented to help the deletion of the CuPy array. 

Together with [this fix in `httomolibgpu`](https://github.com/DiamondLightSource/httomolibgpu/pull/145), it solves the GPU memory leak issue.


dkazanc:
thanks a lot @ptim0626 for this! I've tested on various data sizes 20, 40 and 80Gb. All works as expected and the tests now pass on `iris`.

PR #425, opened 2024-08-22
CI issue fixing + tests

### LINKED ISSUE #427: CI failures for iris tests and docs
IRIS CI tests fail: 

- Pillow [issue](https://github.com/DiamondLightSource/httomo/actions/runs/10511793271/job/29123582872?pr=425#step:8:31) 
- Older CuPy version [installed](https://github.com/DiamondLightSource/httomo/actions/runs/10523783123/job/29159202679?pr=426#step:5:520).

Docs Github CI [fail](https://github.com/DiamondLightSource/httomo/actions/runs/10522588529/job/29155533609?pr=426#step:6:123) on the YAML generator stage because of CuPy import required.

### LINKED ISSUE #431: Broken pipeline tests on `main` due to removal of `auto` param in `remove_outlier` from httomolibgpu
This PR in httomolibgpu https://github.com/DiamondLightSource/httomolibgpu/pull/150 which removed `cucim` changed the `remove_outlier` method in httomolibgpu to not need an `axis` parameter anymore. (This is because the `cucim` method was working on 2D data, but the removal of `cucim` caused the `remove_outlier` and `median_filter` methods to work on 3D data once more - working on 3D data means that there's no axis to iterate over 2D images now).

However, some pipeline files still specify the `axis` parameter for the `remove_outlier` method from httomolibgpu, which causes an error: https://github.com/DiamondLightSource/httomo/blob/bb953914ccd4ead6d52b26b5e8796a099ab78703/tests/samples/pipeline_template_examples/pipeline_gpu1.yaml#L24-L29

Also, the dezinging wrapper assumes that an `axis` parameter is needed, and if it's omitted from the pipeline then the wrapper will automatically insert a default value for an `axis` parameter, which now breaks anytime the `remove_outlier` or `median_filter` methods from httomolibgpu are used (because they no longer have an `axis` parameter): https://github.com/DiamondLightSource/httomo/blob/bb953914ccd4ead6d52b26b5e8796a099ab78703/httomo/method_wrappers/dezinging.py#L51-L55

### LINKED ISSUE #433: Broken pipeline tests due to incorrect stats value assertions
After fixing #431 locally via #432, there are still two failing tests due to failing assertions on stats values:
```
======================================== short test summary info =========================================
FAILED tests/test_pipeline.py::test_run_diad_pipeline_gpu - AssertionError: assert 'Global min
-0.0326580' in '2024-09-05 15:22:42.517 | INFO     | httomo.utils:...
FAILED tests/test_pipeline.py::test_run_pipeline_360deg_gpu2 - AssertionError: assert 'Global
min -0.002859' in '2024-09-05 15:22:45.905 | INFO     | httomo.utils:l...
================================ 2 failed, 15 passed in 65.56s (0:01:05) =================================
```

Possibly this is related to the updated `remove_outlier` method that caused the issue in #431.

dkazanc:
Fixes #427 
Fixes #431 
Fixes #433

## Checklist


httomo/methods_database/packages/external/httomolibgpu/supporting_funcs/prep/phase.py
```diff
@@ -101,7 +101,7 @@ def _calc_memory_bytes_paganin_filter_tomopy(
         + out_slice_size
         + 2 * complex_slice
         + 0.5 * fftplan_slice
-        + 2 * res_slice
+        + res_slice
```
yousefmoazzam, line 104: The memory hook tests are still failing for the paganin method I think, so why has this been changed?

tests/test_backends/test_httomolibgpu.py
```diff
@@ -154,9 +154,9 @@ def test_remove_outlier_memoryhook(flats, ensure_clean_memory, dtype, slices):
 
 
 @pytest.mark.cupy
-@pytest.mark.parametrize("slices", [64, 128])
-@pytest.mark.parametrize("dim_x", [81, 260, 320])
-@pytest.mark.parametrize("dim_y", [340, 135, 96])
+@pytest.mark.parametrize("slices", [2, 4, 8])
+@pytest.mark.parametrize("dim_x", [1024, 2048])
+@pytest.mark.parametrize("dim_y", [1024, 2048])
```
yousefmoazzam, line 159: Why have these been changed?

tests/test_pipeline.py
```diff
@@ -666,9 +666,9 @@ def test_run_diad_pipeline_gpu(
     assert "Path to data: /entry/imaging/data" in verbose_log_contents
     assert "Preview: (100:3101, 8:15, 0:26)" in verbose_log_contents
     assert "Data shape is (3001, 7, 26) of type uint16" in verbose_log_contents
-    assert "Global min -0.0326580" in verbose_log_contents
-    assert "Global max 0.037757" in verbose_log_contents
-    assert "Global mean 0.000327" in verbose_log_contents
+    assert "Global min -0.00915" in verbose_log_contents
+    assert "Global max 0.01134" in verbose_log_contents
+    assert "Global mean 0.00013" in verbose_log_contents
```
yousefmoazzam, line 671: What has caused these to need to be changed?

tests/test_pipeline.py
```diff
@@ -698,9 +698,9 @@ def test_run_pipeline_360deg_gpu2(
     assert "Loading data: tests/test_data/360scan/360scan.hdf" in verbose_log_contents
     assert "Path to data: entry1/tomo_entry/data/data" in verbose_log_contents
     assert "Data shape is (3601, 3, 2560) of type uint16" in verbose_log_contents
-    assert "Global min -0.002859" in verbose_log_contents
-    assert "Global max 0.00561" in verbose_log_contents
-    assert "Global mean 0.000889" in verbose_log_contents
+    assert "Global min -0.00275" in verbose_log_contents
+    assert "Global max 0.005299" in verbose_log_contents
+    assert "Global mean 0.0008" in verbose_log_contents
```
yousefmoazzam, line 703: Ditto: what has caused these to need to be changed?

yousefmoazzam (commented):
Looks good to me :slightly_smiling_face: 

One thing I'd say is that, if multiple things are going to be changed/fixed, could everything that is being changed/fixed be accounted for?

The bits of code I've commented on, I would just like some note in the PR or something that explains the reason for the change, because otherwise some of them look kinda random (like the change in parametrisation values for the paganin memory hook test, or the pipeline stats value assertion changes - is this fixing #433 maybe?).

tests/test_pipeline.py
dkazanc: that is because we use 3d outliers/median instead of 2d now.

tests/test_pipeline.py
dkazanc: same as above, the pipeline does have median or outlier method in it

tests/test_backends/test_httomolibgpu.py
yousefmoazzam: Yep, ok, that's reasonable.

However, I'm wondering if that should be done in the PR that fixes the paganin memory hooks rather than here? Is there a specific reason to put it in this PR (given that the PR is titled "CI issue fixing + tests", while this change doesn't seem to do any CI fixing or test fixing I think)?

dkazanc: sure, I added it back with [291739a](https://github.com/DiamondLightSource/httomo/pull/425/commits/291739addf1e0eff4aaad54ecab902674cd60f38)

Issue #423, opened 2024-08-16
chunk size benchmarks for intermediate files


dkazanc:
investigate how the size of chunks affects the writing speed (reading is not as important in our case as the result is usually saved). From a quick play with the `--frames-per-chunk` parameter, it can be seen the the speed of writing differs significantly. For 40gb raw dataset, saving 54 Gb data. 
```
1 frame per chunk - 28s
4fpc = 13.3s
16fpc = 12.2
32fpc - 14.8s
```
@ptim0626 also mentioned that we should look into a possibility of changing `rdcc` values according to [this](https://docs.h5py.org/en/stable/high/file.html#chunk-cache) guide. In the code, I believe, this is the [place](https://github.com/DiamondLightSource/httomo/blob/bb953914ccd4ead6d52b26b5e8796a099ab78703/httomo/data/dataset_store.py#L238) where it needs to be changed to something like  this for 16 fpc?
```
        self._h5file = h5py.File(
            file,
            "w",
            driver="mpio",
            comm=comm,
            rdcc_nslots=1e5,
            rdcc_nbytes=4 * (2048**2 * 16),
            rdcc_w0=1,
        )
```

dkazanc:
Just to add more info on what I heard from the hdf5 group on NOBUGS about our sizes of chunks when writing the intermediate files. They are suggested for us to try chunks that are not taking the whole frame dimension, but rather a smaller non-rectangular part of it and increase the depth.

Currently we have a parameter `--frames-per-chunk` which will take the whole frame, for instance, for the sinogram case, our one frame will be something of `[1800, 2560]` with sizes are [number of projections, detectorX]. Then if we to increase the depth, this  would result in very large chunks. They are suggested for us to try much smaller chunks with some depth. Following the example above, the chunk can be [180, 256, depth]. And vary the depth in a range [2:32]. 

If this would accelerate the slow writing of sinogram sliced data (e.g., the result of the reconstruction), we might want to rethink the `--frames-per-chunk` parameter and rather specify the size of the chunk relatively, e.g.,  10% of a full frame and the depth?

PR #426, opened 2024-08-22
adding watermark to sweep feature while saving images

### LINKED ISSUE #420: Image masking/stamping with sweep values
A useful feature will be for a user to quickly find the right parameter (e.g. CoR) using `sweep` functionality. "Stamping" each saved image with a value of the sweep parameter.
As we insert `save_to_images`  automatically #419 , we can also insert an additional method that would add a stamp on the data BEFORE the data is being saved as an image. The method itself could be a part of the `httomolib` library. The method should be a bit "smart", so that the numbers will be be visible in different circumstances. I'll add an issue on that in `httomolib`.

dkazanc:
Fixes #420 

## Checklist


httomo/data/dataset_store.py
```diff
@@ -235,7 +235,15 @@ def _create_h5_data(
...
-        self._h5file = h5py.File(file, "w", driver="mpio", comm=comm)
+        self._h5file = h5py.File(
+            file,
+            "w",
+            driver="mpio",
+            comm=comm,
+            rdcc_nslots=1e5,
+            rdcc_nbytes=4 * (2048**2 * 16),
+            rdcc_w0=1,
+        )
```
yousefmoazzam, line 246: May I ask why this has been added in this PR?

I've seen #423 so I'm aware that improving the write speed for the dataset store when it's backed by an hdf5 file is wanting to be explored. However, as far as I can tell, that's a totally separate feature to incorporating the "watermarking" for images saved in a sweep run? (For example, I think that sweep runs don't use the reader/writer in `dataset_store.py`, they use the reader/writer from `param_sweep_store.py`).

So, unless they're connected in some way (please do say so if this is the case, I might have missed something! :sweat_smile: ), I'd suggest that the writing to the dataset store is left alone in this PR and only changes relevant to the watermark incorporation for sweep runs are added here.

dkazanc: ah, good for noticing it. It is accidentality there, needs to be removed. 
I've been playing with those values a bit as @ptim0626 suggested. It seems that increasing the number of chunks larger than 1 (16 seems to be a sweat spot) increases the writing speed of hdf5 intermediate files at least 2 times. So it should be certainly in a different branch, and those values are not optimised anyway.

Issue #358, opened 2024-05-31
Simplify `memory_gpu` field in methods database entry structure


yousefmoazzam:
## Current structure

YAML
```yaml
memory_gpu: 
   - datasets: [tomo] 
   - multipliers: [2.1] 
   - methods: [direct] 
```

where the value of the `memory_gpu` field parses into the following python:
```python
[
  {"datasets": ["tomo"]},
  {"multipliers": [2.1]},
  {"methods": ["direct"]},
]
```

## Goal for simplified structure

YAML
```yaml
memory_gpu: 
  datasets: "tomo"
  multipliers: 2.1
  methods: "direct"
```

where the value of the `memory_gpu` field parses into the following python:
```python
{
  "datasets": "tomo",
  "multipliers": 2.1,
  "methods": "direct",
}
```

- [ ] #356
- [ ] #357

team-gpu:
Also note that [this loop](https://github.com/DiamondLightSource/httomo/blob/659d451c977b9d3bd7cfe17581ce36f3aee61820/httomo/method_wrappers/generic.py#L410) will have to go too. Originally there was the option to have `darks` and `flats` fields there too.

yousefmoazzam:
I wanted to make some comments here, because the mention of "multiple datasets" did raise some questions on my end about how we handle the case when a GPU method requires multiple inputs.

## Example of a method needing multiple inputs

There was one example I remember from a while ago of a method needing multiple inputs, it was a method theoretically being given both:
- data to process
- a mask of some sort

and the mask array was the output of some method.

Because we started httomo with the concept of "named datasets", I believe that one of the ways we had envisioned handling the above example was by:
- naming the input data dataset as something, like `tomo`
- whenever the `tomo` dataset was processed by a method, overwrite the `tomo` dataset with the new result
- when the method that generated the mask completed, name its output dataset so then it could be used later in the pipeline, something like `mask` maybe
- when the pipeline execution gets to the method that needs both the input data (the `tomo` dataset) and the mask (the `mask` dataset), pass both of those named datasets to that method

which is very much like how Savu was used.

## Named datasets not really being used in the runner anymore

However, when refactoring the task runner with Jorg, we came to realise that we didn't really need to name datasets. This was because the output of a method was assumed to overwrite the method's input; this is an implicit assumption with how we process blocks that we are overwriting the input - so why bother relying on naming the input data dataset in the framework if:
- the input dataset to every method is always the output of the previous method
[...]

Issue #357, opened 2024-05-31
Replace unnecessary list of dicts in methods database entries


yousefmoazzam:
An example of the current structure of a methods database entry is the following: https://github.com/DiamondLightSource/httomo/blob/4429711ab89ed1efe86d9c3be260df25fc7f1999/httomo/methods_database/packages/external/httomolibgpu/httomolibgpu.yaml#L3-L11

Focusing on the `memory_gpu` field: https://github.com/DiamondLightSource/httomo/blob/4429711ab89ed1efe86d9c3be260df25fc7f1999/httomo/methods_database/packages/external/httomolibgpu/httomolibgpu.yaml#L8-L11

its values will get parsed into the following data structure in python:
```python
[
  {"datasets": ["tomo"]},
  {"multipliers": [2.1]},
  {"methods": ["direct"]},
]
```

A naive guess would be that it's likely not needed to have a list of dicts, and one single dict like so would suffice:
```python
{
  "datasets": ["tomo"],
  "multipliers": [2.1],
  "methods": ["direct"]
}
```

If so, the YAML structure could be modified to be the following (note the missing leading `-` for each subfield within `memory_gpu`):
```yaml
memory_gpu: 
  datasets: ["tomo"]
  multipliers: [2.1]
  methods: ["direct"]
```

PR #428, opened 2024-08-30
Simplify memory_gpu field in method database entry

### LINKED ISSUE #356: Remove unnecessary single-element lists in methods database entries
An example of the current structure of a methods database entry is the following: https://github.com/DiamondLightSource/httomo/blob/4429711ab89ed1efe86d9c3be260df25fc7f1999/httomo/methods_database/packages/external/httomolibgpu/httomolibgpu.yaml#L3-L11

Focusing on the `memory_gpu` field: https://github.com/DiamondLightSource/httomo/blob/4429711ab89ed1efe86d9c3be260df25fc7f1999/httomo/methods_database/packages/external/httomolibgpu/httomolibgpu.yaml#L8-L11

its values will get parsed into the following data structure in python:
```python
[
  {"datasets": ["tomo"]},
  {"multipliers": [2.1]},
  {"methods": ["direct"]},
]
```

In each dict in the top-level list, there is one key-value pair. The value in each key-value pair is a list with a single element, ie:
```python
["tomo"]
[2.1]
["direct"]
```

A naive guess would be that the list is probably not needed, and would suffice to be parsed into a python data structure as:
```python
[
  {"datasets": "tomo"},
  {"multipliers": 2.1},
  {"methods": "direct"},
]
```

If the list indeed isn't needed, then the YAML can be changed to the following to avoid the list with a single value inside (note the omission of the square brackets):
```yaml
memory_gpu: 
  - datasets: tomo 
  - multipliers: 2.1 
  - methods: direct
```

jessicavers:
Fixes #356, #357, #358

## Checklist


httomo/method_wrappers/generic.py
```diff
@@ -426,25 +426,24 @@ def calculate_max_slices(
...
-            else:
-                (
-                    memory_bytes_method,
-                    subtract_bytes,
-                ) = self._query.calculate_memory_bytes(
-                    non_slice_dims_shape, data_dtype, **self.config_params
-                )
+        subtract_bytes = 0
+        # loop over the dataset names given in the library file and extracting
+        # the corresponding dimensions from the available datasets
```
yousefmoazzam, line 431: Correct me if I'm wrong, but I _think_ this code comment is partially outdated/incorrect (the reference to the now removed loop, and the reference to "datasets")?

If I've got that right, could you either update appropriately or remove the code comment?

httomo/utils.py
```diff
@@ -84,7 +84,9 @@ def log_exception(output: str) -> None:
 
 
 def _parse_preview(
-    preview: List[Dict[str, int]], data_shape: Tuple[int], data_indices: List[int]
+    preview: List[Optional[Dict[str, int]]],
+    data_shape: Tuple[int],
+    data_indices: List[int],
```
yousefmoazzam, line 89: This is a nitpick, but are these changes needed for the PR to do what it's intended to do?

If I remember correctly, the `_parse_preview()` function in `utils.py` is only used by the old loader and not the new one, so I would have thought modifications to this function don't do anything of use?

This looks to be a type-hint improvement which is good, but for a function that is no longer used and unrelated to the rest of the PR.

It of course doesn't hurt to have this change, but I'd like to know if this is needed or not, can you check please?

yousefmoazzam (commented):
Looks good to me, just a few comments :slightly_smiling_face: I'd like the comment about the code comment addressed, but the other one isn't such a big deal, feel free to disregard.

Bigger picture, the test locally are mostly passing but with some failures, due to the `axis` parameter stuff that existed. This issue has recently been fixed, so can you update this branch with the changes from `main` please; that should pick up:
- fixes to IRIS tests CI, so then the IRIS tests run ok
- fixes to docs CI
- updates to fix failing tests due to `axis` issues

yousefmoazzam (approved):
Cool, thanks for this! :slightly_smiling_face:

IRIS tests CI isn't happy only due to paganin memory hook tests, which is planned to get fixed soon in another PR, so this looks good to merge.

PR #432, opened 2024-09-05
Fix tests broken by `axis` param change in httomolibgpu methods

### LINKED ISSUE #431: Broken pipeline tests on `main` due to removal of `auto` param in `remove_outlier` from httomolibgpu

yousefmoazzam:
Fixes #431

## Checklist

### DIFF
--- a/httomo/method_wrappers/dezinging.py
+++ b/httomo/method_wrappers/dezinging.py
@@ -48,11 +48,14 @@ class DezingingWrapper(GenericMethodWrapper):
         # check if data needs to be transfered host <-> device
         block = self._transfer_data(block)
 
-        args = self._build_kwargs(self._transform_params(self._config_params), block)
-        # plug in the correct value for axis instead of auto. In case if axis is absent we take
-        # it equal to 0. This is to avoid failure, but (missing) template parameters should be
-        # tested in the yaml checker potentially
-        self._config_params["axis"] = args.get("axis", 0)
+        if "axis" in self.parameters:
+            args = self._build_kwargs(
+                self._transform_params(self._config_params), block
+            )
+            # plug in the correct value for axis instead of auto. In case if axis is absent we
+            # take it equal to 0. This is to avoid failure, but (missing) template parameters
+            # should be tested in the yaml checker potentially
+            self._config_params["axis"] = args.get("axis", 0)
 
         with catch_gputime() as t:
             block.data = self.method(block.data, **self._config_params)
--- a/tests/samples/pipeline_template_examples/DLS/01_diad_pipeline_gpu.yaml
+++ b/tests/samples/pipeline_template_examples/DLS/01_diad_pipeline_gpu.yaml
@@ -30,7 +30,6 @@
   parameters:
     dif: 0.1
     kernel_size: 3
-    axis: auto
 - method: normalize
   module_path: httomolibgpu.prep.normalize
   parameters:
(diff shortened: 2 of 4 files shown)


yousefmoazzam:
It looks like the IRIS tests + docs CI are kinda messed up right now (ie, #427), so both are reporting errors.

However, I'm pretty sure this change is going to be needed when the IRIS tests CI is eventually fixed: httomolibgpu has recently had a new version v2.1.1, and it [contains](https://github.com/DiamondLightSource/httomolibgpu/commits/v2.1.1/) the `auto` parameter changes that are being addressed on the httomo side in this PR.

httomo/method_wrappers/dezinging.py
```diff
@@ -48,11 +48,14 @@ def execute(self, block: T) -> T:
         # check if data needs to be transfered host <-> device
         block = self._transfer_data(block)
 
-        args = self._build_kwargs(self._transform_params(self._config_params), block)
-        # plug in the correct value for axis instead of auto. In case if axis is absent we take
-        # it equal to 0. This is to avoid failure, but (missing) template parameters should be
-        # tested in the yaml checker potentially
-        self._config_params["axis"] = args.get("axis", 0)
+        if "axis" in self.parameters:
```
dkazanc, line 51: for that one I was quite [explicit](https://github.com/DiamondLightSource/httomo/pull/425/files#diff-dfbcc5d5a18bd1d59d6fdbea5528bdd1f6fdafb6c33d728c026bebfaa0cc3a77R55) that we should be with TomoPy that requires is. But I guess your approach is more general just in case if `axis` is coming back in `httomolibgpu`?

yousefmoazzam: Yeah, I wasn't sure if `axis` could potentially come back in the methods in httomolibgpu, so I did it this way.

By the way, I didn't realise that this issue was being fixed among the other issues in #425, so feel free to close this PR if you'd prefer to fix it there :slightly_smiling_face:

dkazanc: Yeah, sorry things started to misbehave in different places so I got a bit chaotic fixing them, that is why this quite generically named [PR](https://github.com/DiamondLightSource/httomo/pull/425) exists.

dkazanc:
the fix moved to #425

Issue #436, opened 2024-09-11
Direct memory estimators data type change


dkazanc:
Currently direct data estimators do not take into consideration the change of the data type. If the method changes the type, e.g. normalisation, this should be reflected in the library file, e.g., by incorporating the field `output_data_type`. 

For example:
`output_data_type: float32`

In addition, it worth remembering that we convert raw uint16 data to float32 after it gets loaded. Some methods can occur before normalisation which might work with uint16. Worth considering if such a strict conversion really necessary early on.

PR #438, opened 2024-09-11
Fixing OOM issue

### LINKED ISSUE #437: Incorrect memory estimation for median/outlier method
Partially related to #436 issue.

dkazanc:
Fixes #437 

Slightly hacky fix for #437 to deal with the fact that the conversion to float32 happens after the data gets loaded. We should   essentially ignore `source.dtype`.  I wrote more about it in #436, but may be a separate issue should exist to deal with this in future separately from the library changes. 

## Checklist

### DIFF
--- a/httomo/runner/task_runner.py
+++ b/httomo/runner/task_runner.py
@@ -333,8 +333,8 @@ class TaskRunner:
         max_slices_methods = [max_slices] * len(section)
 
         # NOTE: as the convertion of the raw data from uint16 to float32 happens after the data gets loaded,
-        # we should consider self.source.dtype to be float for memeory estimators. This should change later
-        # when the direct memory estimators will be able to tackle the change of data type.
+        # we should consider self.source.dtype to be float for memory estimators. This should change later
+        # when we deal with a possibility of data blocks being of different data type.
 
         # loop over all methods in section
         for idx, m in enumerate(section):
--- a/httomo/runner/task_runner.py
+++ b/httomo/runner/task_runner.py
@@ -301,7 +301,7 @@ class TaskRunner:
 
         data_shape = self.source.chunk_shape
         max_slices = data_shape[slicing_dim]
-
+        
         # loop over all methods in section
         has_gpu = False
         for idx, m in enumerate(section):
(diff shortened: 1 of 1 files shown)


httomo/runner/task_runner.py
```diff
@@ -332,6 +332,10 @@ def determine_max_slices(self, section: Section, slicing_dim: int):
 
         max_slices_methods = [max_slices] * len(section)
 
+        # NOTE: as the convertion of the raw data from uint16 to float32 happens after the data gets loaded,
+        # we should consider self.source.dtype to be float for memory estimators. This should change later
+        # when we deal with a possibility of data blocks being of different data type.
```
yousefmoazzam, line 337: I think it's fine to explain why something is the way it is in a code comment, but I think that anything about what will change in the future is best put into an issue and then referenced in the code comment.

So I think the first part is reasonable to have:
```python
        # NOTE: as the convertion of the raw data from uint16 to float32 happens after the data gets loaded,
        # we should consider self.source.dtype to be float for memory estimators.
```

but I think the next part:
```
This should change later when we deal with a possibility of data blocks being of different data type.
```

shouldn't just be in the source code, I think it's best put in an issue, and then referenced in the code comment.

This is so then work that is to be done in the future isn't documented solely in the source code, but in a github issue or something more easily "visible".

httomo/runner/task_runner.py
```diff
@@ -340,7 +344,7 @@ def determine_max_slices(self, section: Section, slicing_dim: int):
 
             output_dims = m.calculate_output_dims(non_slice_dims_shape)
             (slices_estimated, available_memory) = m.calculate_max_slices(
-                self.source.dtype,
+                np.dtype('float32'), # self.source.dtype, # see the NOTE about the dtype above
```
yousefmoazzam, line 347: This is a nitpick, but this looks a bit funny due to the double `#` (I'm assuming the old code was commented out?).

Also, being a bit more explicit about something that's hardcoded might be useful.

Could you maybe define a variable:
```python
SOURCE_DTYPE = np.dtype('float32')
```
alongside the main comment added at line 335, and then change this line to something like the following, to remove the double `#`:
```python
SOURCE_DTYPE, # see the NOTE about the dtype above
```

httomo/runner/task_runner.py
```diff
@@ -301,7 +301,7 @@ def determine_max_slices(self, section: Section, slicing_dim: int):
 
         data_shape = self.source.chunk_shape
         max_slices = data_shape[slicing_dim]
-
+
```
yousefmoazzam, line 304: Black complains... :sweat_smile:

yousefmoazzam (commented):
This does indeed fix the CUDA OOM error for me when running on a cluster node with P100's :tada: 

Few comments to tidy things up a bit, feel free to take it or leave it.

PR #446, opened 2024-09-12
Integrate padding into runner

yousefmoazzam:
This enables the task runner to load and process padded blocks.

## Checklist

### DIFF
--- a/httomo/loaders/standard_tomo_loader.py
+++ b/httomo/loaders/standard_tomo_loader.py
@@ -390,7 +390,7 @@ class StandardLoaderWrapper(LoaderInterface):
         self.angles = angles
         self.preview = preview
 
-    def make_data_source(self) -> DataSetSource:
+    def make_data_source(self, padding: Tuple[int, int] = (0, 0)) -> DataSetSource:
         assert self.pattern in [Pattern.sinogram, Pattern.projection]
         loader = StandardTomoLoader(
             in_file=self.in_file,
@@ -402,6 +402,7 @@ class StandardLoaderWrapper(LoaderInterface):
             preview_config=self.preview,
             slicing_dim=1 if self.pattern == Pattern.sinogram else 0,
             comm=self.comm,
+            padding=padding,
         )
         (self._angles_total, self._detector_y, self._detector_x) = loader.global_shape
         return loader
--- a/httomo/method_wrappers/generic.py
+++ b/httomo/method_wrappers/generic.py
@@ -1,3 +1,5 @@
+from importlib import import_module
+
 import httomo.globals
 from httomo.block_interfaces import T, Block
 from httomo.runner.gpu_utils import get_gpu_id, gpumem_cleanup
@@ -83,7 +85,6 @@ class GenericMethodWrapper(MethodWrapper):
         self._comm = comm
         self._module_path = module_path
         self._method_name = method_name
-        from importlib import import_module
 
         self._module = import_module(module_path)
         self._method: Callable = getattr(self._module, method_name)
--- a/httomo/method_wrappers/save_intermediate.py
+++ b/httomo/method_wrappers/save_intermediate.py
@@ -70,13 +70,15 @@ class SaveIntermediateFilesWrapper(GenericMethodWrapper):
         # - we return just the input as it is
         self._gpu_time_info = GpuTimeInfo()
         with catchtime() as t:
-            data = block.data if block.is_cpu else xp.asnumpy(block.data)
+            data = (
+                block.data_unpadded if block.is_cpu else xp.asnumpy(block.data_unpadded)
+            )
         self._gpu_time_info.device2host += t.elapsed
 
         self._method(
             data,
             global_shape=block.global_shape,
-            global_index=block.global_index,
+            global_index=block.global_index_unpadded,
             slicing_dim=block.slicing_dim,
             file=self._file,
             frames_per_chunk=httomo.globals.FRAMES_PER_CHUNK,
--- a/httomo/methods_database/packages/external/httomolibgpu/httomolibgpu.yaml
+++ b/httomo/methods_database/packages/external/httomolibgpu/httomolibgpu.yaml
@@ -14,7 +14,7 @@ misc:
       output_dims_change: False
       implementation: gpu_cupy
       save_result_default: False
-      padding: False
+      padding: True
       memory_gpu:
         multiplier: 2.1
         method: direct
--- a/httomo/methods_database/packages/external/httomolibgpu/httomolibgpu.yaml
+++ b/httomo/methods_database/packages/external/httomolibgpu/httomolibgpu.yaml
@@ -5,7 +5,7 @@ misc:
       output_dims_change: False
       implementation: gpu_cupy
       save_result_default: False
-      padding: False
+      padding: True
       memory_gpu:
         multiplier: 2.1
         method: direct
--- a/httomo/methods_database/packages/external/httomolibgpu/supporting_funcs/misc/corr.py
+++ b/httomo/methods_database/packages/external/httomolibgpu/supporting_funcs/misc/corr.py
@@ -1,0 +1,11 @@
+from typing import Tuple
+
+
+__all__ = [
+    "_calc_padding_remove_outlier",
+]
+
+
+def _calc_padding_remove_outlier(**kwargs) -> Tuple[int, int]:
+    kernel_size = kwargs["kernel_size"]
+    return (kernel_size // 2, kernel_size // 2)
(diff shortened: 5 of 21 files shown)


yousefmoazzam:
@dkazanc Only one method, `remove_outlier`, that can make use of the padding feature has been enabled so far, purely as a means of testing. Feel free to add support for any other methods as a means of testing if the padding feature works or not :slightly_smiling_face:

dkazanc:
Thanks Yousef, can you please check if you get an assertion error by running the following pipeline. Note the saving is enabled after the median filter, if the saving is disabled the pipeline works. I'll add the median enabled padding bit in a second.  cheers

```
- method: standard_tomo
  parameters:
    name: tomo
    rotation_angles:
    preview:
      detector_y:
        start: 900
        stop: 1120
- method: find_center_vo
  module_path: httomolibgpu.recon.rotation
  parameters:
    ind: mid
    smin: -50
    smax: 50
    srad: 6
    step: 0.25
    ratio: 0.5
    drop: 20
  id: centering
  side_outputs:
    cor: centre_of_rotation
- method: normalize
  module_path: httomolibgpu.prep.normalize
  parameters:
    cutoff: 10.0
    minus_log: true
    nonnegativity: false
    remove_nans: true
- method: median_filter
  module_path: httomolibgpu.misc.corr
  parameters:
    kernel_size: 5
    dif: 0.0
  save_result: true
- method: FBP
  module_path: httomolibgpu.recon.algorithm
  parameters:
    center: ${{centering.side_outputs.centre_of_rotation}}
    filter_freq_cutoff: 1.1
    recon_size: null
    recon_mask_radius: null
  save_result: true
```

dkazanc:
OK, I've looked at the result of the padding. Here is the obvious poorly stitched region of unpadded data after median filter with kernel 7.

And here is the padded one (for now I ignored the first block and the last block not saved due to the error above). Interestingly the sharp line has reduced its intensity but not fully removed, something remains there. I guess the reason that the line is in the different place is that the block size has changed. Still I wonder why there is a line left...  I'll check the padding calculation, may be we padded slightly less than needed.

yousefmoazzam:
Oh yeah I see what you mean, I have to squint a bit at the second image, but I do see a fainter line still!

I'll have a think/look and see if I can come up with anything to explain this unexpected behaviour - looks like the padding is sort of half working only... :sweat_smile:

yousefmoazzam:
Also, the latest commit I pushed _should_ have fixed the error when saving the last block.

dkazanc:
It could be an issue of an index, i.e., the padded region is slightly smaller (1 pixel smaller) than it should be. So I went to `_calc_padding_median_filter` and increased the padding region to 1 like this `(kernel_size // 2 + 1, kernel_size // 2 + 1)`. It seems that it solved the problem, I see not boundaries here. So may be worth rechecking the indices quickly? And yes the data is saved in full now, thanks. 
padded2

dkazanc:
It would be nice to have a test that runs a method with padding on the whole region as a full 3D method and then also  imitates the blocking with more processed padded blocks than needed. The the framework does it thing to unpad, etc., but the results of both approaches should be identical. 
E.g., the result of full-data 3D median = the result of padded N-number of blocks 3D median. 
For some methods, e.g., iterative ones, this won't hold, but for median it should.

yousefmoazzam:
> It would be nice to have a test that runs a method with padding on the whole region as a full 3D method and then also imitates the blocking with more processed padded blocks than needed. The the framework does it thing to unpad, etc., but the results of both approaches should be identical. E.g., the result of full-data 3D median = the result of padded N-number of blocks 3D median. For some methods, e.g., iterative ones, this won't hold, but for median it should.

Mmm, I had a similar thought when you shared the first pair of screenshots. It feels like, to have absolute confidence that the both the following produce the same result:
- without padding, processing an entire 3D chunk in one go
- with padding, processing two 3D blocks from the chunk sequentially

a test making assertions on arrays would be best, so then we don't have to rely on visual inspection.

## Idea for a test

I was having a think about at what "level" would be the best place to test this. The objects that are involved in padded blocks have tests, and we need to be careful to not have this new test testing stuff that is already tested.

My idea after an initial thought was that this could be a test for the task runner:
- give it a section that contains a single method (to keep it as simple as possible), one that needs padding
- execute that single section, and force it to have multiple blocks, so we have multiple padded blocks being processed
  - the `_execute_section()` method is private, which isn't the nicest to run manually, but it seems like the easiest way currently, and other task runner tests do this too so maybe it's good enough for the moment? :shrug: 
- run the method separately on the same input, but as one full array, not split into multiple blocks
- compare the padded output execution saved in the sink to what running the method separately does
    -  the data in the sink has only the core part of the processed blocks are written to the sink, the padded slices are thrown away, so it in theory should be the same as when running the method on the entire chunk
  - the `.sink` attribute on the task runner is public so it's reasonable to access it

It would then be considered an integration test of the runner for padding I guess, which doesn't yet exist I think. And it avoids testing padding for the lower-level objects (like `DataSetBlock` or the wrappers), which already have padding tests.

## 3D method requiring padding that is used in the test

Another thing to mention is that, if possible, I'd prefer to not have to rely on a 3D method in some external package for testing padding. This is so then the test of the padding framework in httomo is independent of any methods backend.

I'm not 100% sure if it'd work nicely, but if it did, would you be okay with the test defining a small 3D GPU method local to the test, that can be used? The wrapper in the section probably can be patched to use this dummy 3D method, and I reckon all the other stuff relating to the methods database can be fiddled with/mocked in order to get the wrapper in the test to execute this dummy 3D method?

dkazanc:
Yes, that sounds like a good solution. We don't need a 3D method for that indeed. We can simple randomly generate the 3D data first and the method can modify the data, for instance scaling or adding a number. If things will go wrong with the padding, I believe, we should be able to trace it only with that. But the input data must be random or the standard data we use.

yousefmoazzam:
Ok cool, I'll get started on that test :slightly_smiling_face:  (And it should fail at first, given the horizontal lines in the data shown earlier).

One thing I'd like to clarify since I'm not sure now: I thought the dummy method in the test _does_ have to be 3D, in the sense that it has to use the neighbourhood/padded slices in some way? I thought that if it doesn't use the padded slices, then there won't be any difference between:
- running the dummy method with the correct padded blocks
- running the dummy method without padded blocks or incorrect padded blocks

The goal of the test is to be able to tell if the blocks have been padded correctly by the task runner, so I thought we need to be able to see the difference between those two cases? And there's only a difference between the two cases when the method is 3D? Or have I gotten things mixed up (which is quite possible, apologies)?... :sweat_smile:

dkazanc:
thanks @yousefmoazzam . Yes, I think by simply changing a value per pixel won't change anything as I think the stitching of the boundaries back to the original position is happening correctly at the moment (we do not see jumps in the image). So we need a 3D change indeed. The simplest could be a sum over 6+1(central) pixels if we consider `3 x 3 x 3` neighbourhood or 24+1 pixels for `5 x 5 x 5` (padding 2). Could you do that? That should be enough I believe?

yousefmoazzam:
Yeah I agree about the emphasis on having a simple 3D method in the test; while the method does need to be 3D, it doesn't need to be complex, just enough to ensure that a change of neighbourhood has a detectable effect on its output. What you suggested sounds like a good candidate, I'll give it a go :slightly_smiling_face:

dkazanc:
I can confirm that after the fix [0a3cd1a](https://github.com/DiamondLightSource/httomo/pull/446/commits/0a3cd1aae3ed5ea7fd1b310131373dd36750090d) the lines disappeared. So visually all seem to work correctly and I hope that the test should pass.

yousefmoazzam:
Ok, apologies for being quiet after the integration test was added, I was investigating the two issues that are linked above:
1. The median filter issue _appears_ to possibly not be related to the padding, but something specific to the median filter method.
2. The chunk being split into two blocks even if the chunk can fit into GPU memory doesn't feel like it's too urgent, since full chunks often don't fit into GPU memory when running on full data, but obviously is nice to fix for the sake of the number of blocks being produced is as expected.

I think the integration test is mostly ok, but I have added two notes marked `NOTE` that flag the bad things being done (making assumptions about the internals of the task runner).

If we feel that it's a bad way to test things we can always remove the test and try to think of better ways to do it (which probably requires some refactoring to get rid of the need to inspect internals of the task runner), since the horizontal lines were gone even before I wrote the test.

Let me know what you think @dkazanc :slightly_smiling_face:

yousefmoazzam:
Yeah, I think #453 affects small data mostly. I agree that it can be left for later, so I'll merge this to wrap up the main integration of padding into the runner.

PR #454, opened 2024-09-20
Improve paganin filter method's memory estimation

### LINKED ISSUE #430: Fixing memory estimators for Paganin
Memory estimators for Paganin are [failing](https://github.com/DiamondLightSource/httomo/actions/runs/10595929644/job/29362752194) now, after the corrective changes made in `httomolibgpu`.  
The reason so far is purely speculative, the plan could be released before the function exists. Actually tests pass on my machine so the reason could be in incorrect calculation of a plan. The plan size calculation needs to be done in the [same](https://github.com/DiamondLightSource/httomo/blob/bb953914ccd4ead6d52b26b5e8796a099ab78703/httomo/methods_database/packages/external/httomolibgpu/supporting_funcs/recon/algorithm.py#L95) manner as for FBP.

yousefmoazzam:
Fixes #430

## Checklist

### DIFF
--- a/httomo/methods_database/packages/external/httomolibgpu/supporting_funcs/prep/phase.py
+++ b/httomo/methods_database/packages/external/httomolibgpu/supporting_funcs/prep/phase.py
@@ -106,6 +106,9 @@ def _calc_memory_bytes_paganin_filter_tomopy(
     # Size of cropped/unpadded + cast to float32 result of 2D IFFT
     cropped_float32_res_slice = np.prod(non_slice_dims_shape) * np.float32().nbytes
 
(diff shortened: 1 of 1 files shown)


yousefmoazzam:
Ok, both paganin filter method's memory hook tests now pass on:
- IRIS CI
- my local workstation `pc0074`
- a V100 node
- a P100 node

so I'm hopeful that at least the bulk of the significant memory allocations in both methods are accounted for correctly in their respective estimators.

Before I mark this as ready to review I will write up an issue on the changes in b32a2c8, because they will surely raise some questions and thus require a bit of an explanation...

PR #457, opened 2024-09-26
Update pipelines to match changes to image saver and rescaler methods

yousefmoazzam:
These changes are needed on the httomo side to accommodate when the following changes are released in httomolib and httomolibgpu (ie, when these changes make their way into the conda packages that httomo installs during CI):
- https://github.com/DiamondLightSource/httomolibgpu/pull/157
- https://github.com/DiamondLightSource/httomolib/pull/17

## Checklist

### DIFF
--- a/httomo/transform_layer.py
+++ b/httomo/transform_layer.py
@@ -92,8 +92,6 @@ class TransformLayer:
                         task_id=f"saveimage_sweep_{m.task_id}",
                         subfolder_name="images_sweep_" + str(m.method_name),
                         axis=1,
-                        perc_range_min=5,
-                        perc_range_max=95,
                     ),
                 )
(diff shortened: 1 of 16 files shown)


yousefmoazzam:
The two failing tests are both related to how the image saver wrapper is auto-inserted after the method that has the sweep:
```
test_pipeline.py::test_run_gpu_pipeline_sweep_cor
test_pipeline.py::test_run_gpu_pipeline_sweep_paganin
```

The tests fail because the image saver method has been auto-inserted, but the input data has not been rescaled (the image saver no longer supports rescaling the input). Thus, it would seems like, for sweeps, the `rescale_to_int` method needs to be auto-inserted along with the `save_to_images` method to fix this issue.

PR #471, opened 2024-10-07
Remove template generation as a part of the CI

namannimmo10:
## Checklist


namannimmo10:
Since we have https://github.com/DiamondLightSource/httomo-backends as a separate dependency we don't have to generate the templates here.

Issue #492, opened 2024-10-16
Combining `|` and `xp.ndarray` in a type causes warning in sphinx API docs build


yousefmoazzam:
As part of the warnings seen in #468, with a bit of trial and error I have narrowed down the cause of the "unsupported operand `|`" errors being produced by sphinx, despite running with a python 3.12 interpreter (which should recognise the `|` operator as another way to do `Union`).

It seems like mixing `|` and types that come from the `xp` variable causes the issue. If `xp` is not used, then the `|` poses no problem for sphinx.

I found this by being able reproduce the warning sphinx generates, but in a dummy module I created locally in my httomo repo and generating API docs for that module only.

The dummy module (which I lovingly called `bloop.py`) is the following:
```python
from typing import TypeAlias, Union

import numpy as np

try:
    import cupy as xp

    try:
        xp.cuda.Device(0).compute_capability
        gpu_enabled = True  # CuPy is installed and GPU is available
    except xp.cuda.runtime.CUDARuntimeError:
        import numpy as xp

except ImportError:
    import numpy as xp


Blah: TypeAlias = Union[np.ndarray, xp.ndarray]
Thing: TypeAlias = int | float
```

where you can see that I have defined `xp` as it is defined in `utils.py`: https://github.com/DiamondLightSource/httomo/blob/f619ed123f2c5aad790ad75751d737c817a44196/httomo/utils.py#L11-L21

Having `Blah` defined as:
```python
Blah: TypeAlias = Union[np.ndarray, xp.ndarray]
```

causes no issues, but having it defined as:
```python
Blah: TypeAlias = np.ndarray | xp.ndarray
```

produces the warning:
```
(/dls/science/users/twi18192/conda-envs/docs-httomo-py310-updated) [twi18192@pc0074 httomo (sphinx-fixes)]$ source docs/sphinx-build.sh
***********************************************************************
                Starting the sphinx script
***********************************************************************
            Creating plugin API files and html pages ..
Running Sphinx v8.1.3
loading translations [en]... done
making output directory... done
Blah is int | float
[...]

PR #494, opened 2024-10-17
Fix docs build warnings and generate API docs with `autosummary` extension

### LINKED ISSUE #486: API docs built in docs build script, but not included in `index.rst`
From finding out #487, I then wondered where exactly the API docs were in the documentation in https://diamondlightsource.github.io/httomo/, and I couldn't find them.

They're definitely being built, it's in the docs build script: https://github.com/DiamondLightSource/httomo/blob/18c2e044c0a3c5e3dd831ab7139b1676ba5599bb/docs/sphinx-build.sh#L27-L28

but I don't see the built `api/` dir being included in the `index.rst` anywhere https://github.com/DiamondLightSource/httomo/blob/18c2e044c0a3c5e3dd831ab7139b1676ba5599bb/docs/source/index.rst?plain=1

Question 1: Am I missing something, or are we not including the API docs in the "final docs"?

Question 2: If we're not including the API docs, do we want to?

Question 3: If we don't want to include the API docs, can we get rid of the build of them, as one way to get rid of the warnings? (Alternatively, keep the API docs, but fix the short underline warnings?)

### LINKED ISSUE #487: "Title underline too short" warning in a lot of API docs
From browsing the various warnings in the docs build (noted in #468), I noticed these types of warnings for basically every module in httomo:
```
:mod:`httomo.sweep_runner.param_sweep_yaml_loader`
============================================
/dls/science/users/twi18192/httomo/docs/source/api/httomo.sweep_runner.side_output_manager.rst:2: WARNING: Title underline too short.
```

It seems this is due to the hardcoded length of the underline in the template used to generate the API docs rst files from*: https://github.com/DiamondLightSource/httomo/blob/18c2e044c0a3c5e3dd831ab7139b1676ba5599bb/docs/source/_templates/module.rst_t#L1-L7

*The `-t` flag in the docs build script is providing that template I think: https://github.com/DiamondLightSource/httomo/blob/18c2e044c0a3c5e3dd831ab7139b1676ba5599bb/docs/sphinx-build.sh#L28

### LINKED ISSUE #490: Manual mocking of dependencies in docs env, rather than using a sphinx option designed for this purpose
In `conf.py` for the docs, there's a part that's mocking dependencies of httomo that are required during the `autosummary` extension's importing of modules: https://github.com/DiamondLightSource/httomo/blob/f619ed123f2c5aad790ad75751d737c817a44196/docs/source/conf.py#L37-L53

However, there appears to be at least two sphinx options that are designed for this purpose:
- [`autosummary_mock_imports`](sphinx-doc.org/en/master/usage/extensions/autosummary.html#confval-autosummary_mock_imports)
- [`autodoc_mock_imports`](https://www.sphinx-doc.org/en/master/usage/extensions/autodoc.html#confval-autodoc_mock_imports)

It's likely preferable to use the sphinx options to take care of this, since they may be doing other things in the mocking process that our manual way isn't doing.

yousefmoazzam:
Fixes #486 
Fixes #487 
Fixes #490 
_Doesn't_ solve #492, but instead finds a workaround via not using `|` in conjunction with certain types

## Checklist


yousefmoazzam:
All warning in the docs build locally are gone, but the CI has one warning:
```
/home/runner/work/httomo/httomo/docs/source/developers/api.rst:4: WARNING: autosummary: failed to import httomo.cli.
Possible hints:
* AttributeError: module 'httomo' has no attribute 'cli'
* ImportError: cannot import name '__version__' from 'httomo' (/home/runner/work/httomo/httomo/httomo/__init__.py)
* ImportError: 
```
Since it appears in CI only and not locally, it may need a bit more time to debug, so I'll make an issue for it to tackle after this PR; it's small in comparison to the main aim of this PR which was to:
- get rid of as many easily tackle-able warnings in the docs build as possible (ie, get all the low hanging fruit)
- generate API docs and include them in the main `index.rst`

Issue #455, opened 2024-09-23
An option to pass preview to the method


dkazanc:
For distortion correction method, we also need to apply [previewing](https://github.com/DiamondLightSource/httomolibgpu/blob/c4fd8cca024bb4737aa9e6c2a3c2eec7be0f9185/httomolibgpu/prep/alignment.py#L66)  in order to compensate in coefficients for the cropped data. 

At this point I don't know what exactly we need to do in order to pass `preview` to the method. 
One possible choice is to use a side output to the `preview` dictionary in the loader, e.g. 
`${{loader.side_outputs.preview}}` as a part of distortion correction template. Would that be possible? We probably then need to attach `id` to the loader. 
Another option is to "secretly" pass the preview value into the method behind the scenes and hide `preview` parameter in the template. This would need some modifications in the wrappers. But being more explicit here is probably a better way?

yousefmoazzam:
> One possible choice is to use a side output to the `preview` dictionary in the loader, e.g. `${{loader.side_outputs.preview}}` as a part of distortion correction template. Would that be possible? We probably then need to attach `id` to the loader. 

Unfortunately, I think this particular option can't be done without some changes to the loader. Side outputs are a thing that method wrappers can handle, but neither the loader `StandardTomoLoader` nor the loader wrapper `StandardLoaderWrapper` is a method wrapper*, so loader's in general currently don't have functionality to produce side outputs.

This doesn't mean it's not an option, just wanted to point out that it's probably more complicated than it initially seems.

*FYI, when I say the loader isn't a method wrapper, I mean that neither `StandardTomoLoader` nor `StandardLoaderWrapper` implement methods defined on the `MethodWrapper` protocol, like `get_side_outputs()`: https://github.com/DiamondLightSource/httomo/blob/351e028a5298bc0b619eab7280c6b07d2dccd52c/httomo/runner/method_wrapper.py#L175-L179

dkazanc:
what do you think is preferable in this case explicit or implicit? In principle, `preview` argument might need to go into other methods at some point. For instance, the Vo centering method relies on the single index given to the method itself. I think if `mid` is given or a number, we calculate such index with respect to the global data, not the previewed subset. I'm not 100% sure that this is the right way as may be it should be relative to the preview `mid` instead. If that is the case, then we need `preview` parsed into the method. Other methods might also needed at some point, e.g., I was thinking about the autocropping.

So, do you think it would be nicer to implement it more explicit, i.e., via side output?

[...]

PR #507, opened 2024-10-22
Add distortion correction wrapper

### LINKED ISSUE #38: Pass `preview` info to method functions that need it
For example, the distortion correction method can be modified to take into account the `preview` value (as demonstrated in the Savu implementation https://github.com/DiamondLightSource/Savu/blob/master/savu/plugins/corrections/distortion_correction.py#L92).

At the moment, there is no mechanism in the task runner for allowing methods which require `preview` information on their input data that was set in the loader.

For methods which don't have their own `preview` parameter but need `preview` info from the loader, one possible option could be to make `preview` a "hidden parameter" and have some function in the task runner check for method functions which have a "hidden" `preview` parameter as opposed to a "normal" `preview` parameter.

yousefmoazzam:
Fixes #38 (also see #455 for other relevant details)

## Checklist

### DIFF
--- a/httomo/method_wrappers/__init__.py
+++ b/httomo/method_wrappers/__init__.py
@@ -1,3 +1,4 @@
+from httomo.preview import PreviewConfig
 from httomo.runner.method_wrapper import MethodWrapper
 from httomo.runner.methods_repository_interface import MethodRepository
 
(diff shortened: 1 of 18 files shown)


yousefmoazzam:
Note that the change of the distortion correction parameters in https://github.com/DiamondLightSource/httomolibgpu/pull/161 haven't been put into a release yet.

To work with the changes in this PR, the distortion correction memory hook tests below should be updated to reflect these parameter changes: https://github.com/DiamondLightSource/httomo/blob/6fcf035c79836ba09db9b12651e6b0496d21302b/tests/test_backends/test_httomolibgpu.py#L242-L246

The only reason these tests still pass in CI is due to the CI installing the latest version of httomo that is available, which _doesn't_ contain the changes to the distortion correction parameters.

PR #508, opened 2024-10-23
adding extraction of angles related to data indices only

### LINKED ISSUE #506: Rotation angles cropping according to keys
We need to crop the metadata extracted rotation angles so that they correspond to the given keys. Also check the 360 data cropping case #236

dkazanc:
Fixes #506

## Checklist


yousefmoazzam (approved):
Cool, that's nice that the fix was small, looks good to me :slightly_smiling_face:

Issue #504, opened 2024-10-21
Revaluate whether `StandardLoaderWrapper` is needed or not


yousefmoazzam:
## TL;DR

- the runner wants to be able to get a loader generically, so it relies on any implementor of `LoaderInterface` to get one
- the `StandardLoaderWrapper` class implements `LoaderInterface`, so the runner uses `StandardLoaderWrapper` to get a loader
- the `StandardLoaderWrapper` class wraps the `StandardTomoLoader` class, so then the latter can be used by the runner
- the `StandardTomoLoader` class can be very slightly modified (adding three property-getters, `detector_x`, `detector_y`, `angles_total`) to implement `LoaderInterface`, to avoid the extra layer of `StandardLoaderWrapper`
- remove the `make_data_source()` method on the `LoaderIntarface` protocol (which then means that `StanardTomoLoader` now would implement `LoaderInterface`)
- then, make the loader factory function `make_loader()` return an instance of `StandardTomoLoader` instead of `StandardLoaderWrapper`

## Details

The `Pipeline` object represents the loader and methods in the pipeline. In particular, for the loader, it requires anything that implements the `LoaderInterface` protocol: https://github.com/DiamondLightSource/httomo/blob/f6bb8aca8a4fb10cce5dd55c94dff2d0f1d77325/httomo/runner/pipeline.py#L12-L14

When the UI layer creates a `Pipeline` object, it creates an implementor of `LoaderInterface` by using the `make_loader()` function (which could be viewed as a factory function for implementors of `LoaderInterface`): https://github.com/DiamondLightSource/httomo/blob/f6bb8aca8a4fb10cce5dd55c94dff2d0f1d77325/httomo/ui_layer.py#L143-L159

The task runner then uses the `make_data_source()` method defined on the `LoaderInterface` protocol to generate an implementor of `DataSetSource`: https://github.com/DiamondLightSource/httomo/blob/f6bb8aca8a4fb10cce5dd55c94dff2d0f1d77325/httomo/runner/task_runner.py#L226

Now, the way that `StandardLoaderWrapper` implements `LoaderInterface` is _mostly_ by using the `StandardTomoLoader.global_shape` getter: https://github.com/DiamondLightSource/httomo/blob/f6bb8aca8a4fb10cce5dd55c94dff2d0f1d77325/httomo/loaders/standard_tomo_loader.py#L394-L421

Note that the three property getters:
- `detector_x`
- `detector_y`
- `angles_total`

[...]

PR #521, opened 2024-11-04
Paganinsweep

### LINKED ISSUE #424: Sweep issue for smaller previews
As we previously discussed, it was sensible to limit the size of the preview to 7 slices so that the users would avoid running `sweep` pipelines for big datasets. 

I've noticed that for small and larger previews, Paganin filter gives different results for the **_same_** set of parameters.  This means that parameters found for smaller previews cannot be translated to full volume datasets. I suspect that the reason for this is the Fourier filtering nature of the filter which might behave adaptively based on the frequencies present in the data. And since the data is significantly cropped, it is not representative of the whole dataset. This is just a hypothesis, but the fact remains that the result is different, see the images bellow:

Preview of 20 slices
image

Preview of 200 slices
image

So I'm suggesting here to extend the strict 7 slices criteria for `preview` when the `sweep` run is terminated with:
`ValueError: Parameter sweep runs support input data containing <= 7 sinogram slices, input data contains 100 slices`

**Two options:**
1. Catching if the method works with projections and potentially increase the limit to 200-300 slices. May be just specifically for Paganin? But I have a slight concern here about the distortion correction method which might also need a larger preview or even the whole projection.
2. Apply the larger limit globally. This has got the drawback if the method works with sinograms, like CoR, then there is no need for so many slices. 

A question, I believe that `sweep` run currently bypasses memory estimators and blocks?

dkazanc:
Fixes #424

## Checklist


dkazanc:
While working on this I thought that may be we could resolve the limits on the sweep preview defined by the GPU memory by using the CPU Paganin method instead?

In this case, we can be less conservative on what size of the preview we take, even if the whole projection image should be used for a very wide kernel. Also doing things like [this](https://github.com/DiamondLightSource/httomo/blob/7c3ee8f4298b5676fe87af3521e23c5e561ed6b3/httomo/sweep_runner/param_sweep_runner.py#L66) doesn't feel very elegant and the main question how `vertical_slices_preview_max` can be estimated across different GPU devices. To what memory size we should be linking that value? It just feels a bit fragile at the moment. We could, potentially, swap the backend from `httomolibgpu` to `httomolib` for that method? It, however, can fail on some other GPU function, if we decide to take the previewed block size too large. 

Worth having a chat about it @yousefmoazzam ?

yousefmoazzam:
Yeah sure, I'm happy to discuss things at some point.

For now, I'd like to try and enumerate the different points raised for the sake of clarity, I think I counted three (but do correct me if there's more/less):
1. the idea of using the CPU paganin filter in sweep pipelines that involve paganin, to try and reduce the threshold on the maximum number of vertical slices
2. _if_ we keep GPU paganin filter in sweep pipelines: the question of how can the calculation of the maximum vertical slices be made to work with multiple GPUs (ie, how to make the maximum vertical slices threshold generic across GPU models)
3. regardless of if the paganin filter is CPU or GPU: the concern that the modifications to the preview to accommodate the paganin filter kernel size could cause GPU OOM errors for _other_ GPU methods in the sweep pipeline

dkazanc:
thanks @yousefmoazzam, you've summarised the issues well. So I'd say we either should go **full** CPU or GPU, no hybrid modes, as it will essentially result in the same OOM GPU issue at some point. 

Of course the full CPU will make it slightly inconvenient to users as they would need to build a TomoPy analogue. As a possible solution (if we decided to go that way) is to create `httomolib` CPU duplicates of GPU methods in the `httomolibgpu` library.  So that when the user would run the sweep pipeilne with Paganin involved, we would modify the module paths from `httomolibgpu` to `httomolib` leaving everything else intact. Some `httomolib` methods could import and reuse `tomopy` methods, if needed. 

I know that it doesn't sound ideal, but it is one possible way to avoid potentially frequent situations when the large blocks won't fit the GPU resulting in constant warning messages to users. 

if we're still deciding to proceed with the GPU implementation, I think we need to deal with the situation that the largest block [defines](https://github.com/DiamondLightSource/httomo/blob/7c3ee8f4298b5676fe87af3521e23c5e561ed6b3/httomo/sweep_runner/param_sweep_runner.py#L352) the needed size for the sweep run. I suggest that the accepted blocks (bellow the upper limit) will be taken and therefore the list of parameters will be modified accordingly. For instance alpha = [0.1, 0.2, 0.3] is given but the blocks sizes are acceptable only for [0.1, 0.2] values. I'd discard 0.3 in that instance and proceed with the run for 2 parameters, rather than completely abort the whole run.   So you can see, more hacks basically to make the GPU stuff to work...

dkazanc:
OK, so as a conclusion of our discussion @yousefmoazzam , I'll do the following. 

- [ ] get a simple memory estimator that is linearly projecting the amount of slices that fit the GPU for Paganin method based on the GPU available (e.g., 4gb,6gb,12gb,16gb,32gb)  
- [ ] reduce the span of given parameters for Paganin (e.g. `alpha`) if parameters lead to the size of the block larger that is accepted for the current device. If all parameters lead to a kernel that is too wide, kill the run.

yousefmoazzam:
Sounds good, worth a shot to see how it goes :+1: 

It's worth pointing out since we didn't note it in the discussion earlier: with this approach we've addressed points 1 and 2 (in the points listed above in my attempted summary), but point 3 is still unaddressed. Dunno if you'd like more discussion before going ahead and trying this approach, or we just deal with point 3 later, it's up to you.

dkazanc:
Actually to some degree the point 3 is addressed by relying on a memory hungry method (Paganin) to decide the maximum size for a block. I'm hoping that methods in pre/post Paganin pipeline require less memory than Paganin itself. We will see if this is the case when we implement our hack around memory estimation.

dkazanc:
I think I need some help fixing [this](https://github.com/DiamondLightSource/httomo/blob/46e77f97a896926db9908d9cea9ed81d2758e7ba/tests/sweep_runner/test_param_sweep_runner.py#L379) test please @yousefmoazzam . The reason why this test fail is that the `source._data` object [here](https://github.com/DiamondLightSource/httomo/blob/46e77f97a896926db9908d9cea9ed81d2758e7ba/httomo/sweep_runner/param_sweep_runner.py#L163)  is not accessible (mocked?) in the test. 
I'm not sure if there any other way to obtain the unpreviewed shape of the raw data, that is what `source._data.shape[1]`does, or may be the tests itself needs to be modified?

I've tested this approach for 20,40,80Gb datasets (different number angles) and also different GPU cards, mainly P100 and V100. I didn't get OOM error so far. This [factor](https://github.com/DiamondLightSource/httomo/blob/46e77f97a896926db9908d9cea9ed81d2758e7ba/httomo/sweep_runner/param_sweep_runner.py#L343) defines  the number of slices essentially for Paganin method not to break (or any other method in the pipeline I hope). I think the second memory hungry one is FBP so far, it doesn't break. But with Fourier recon we might want to reconsider and increase `factor`. thanks

dkazanc:
And a follow-up, a slightly different approach to the one I suggested earlier. I do not kill the run even if the kernel width is larger than the block that fits the memory. I just take the largest block in this case and proceed with the run.  The users still get the result that is smoothed and probably discard it themselves, but we're safer here from questions why the runs are so frequently terminated.

yousefmoazzam:
I saw that the PR is now marked as ready for review, so I'm happy to go over it at some point soon :slightly_smiling_face: 

For now, I'll try to answer this:

> I think I need some help fixing [this](https://github.com/DiamondLightSource/httomo/blob/46e77f97a896926db9908d9cea9ed81d2758e7ba/tests/sweep_runner/test_param_sweep_runner.py#L379) test please @yousefmoazzam . The reason why this test fail is that the `source._data` object [here](https://github.com/DiamondLightSource/httomo/blob/46e77f97a896926db9908d9cea9ed81d2758e7ba/httomo/sweep_runner/param_sweep_runner.py#L163) is not accessible (mocked?) in the test. I'm not sure if there any other way to obtain the unpreviewed shape of the raw data, that is what `source._data.shape[1]`does, or may be the tests itself needs to be modified?

Yep, the issue seems to be that:
- the test is using a mock loader wrapper + mock loader (to make things simpler and make the test faster, where we don't have to create a real loader that loads an actual file from the filesystem)
- the mock loader has type `DataSetSource` (as expected; that's the protocol for a data source)
https://github.com/DiamondLightSource/httomo/blob/46e77f97a896926db9908d9cea9ed81d2758e7ba/tests/testing_utils.py#L69-L102
- the `DataSetSource` type says nothing about having the private attribute `._data` (as expected; protocols shouldn't enforce what private attributes are needed for an implementation)
https://github.com/DiamondLightSource/httomo/blob/46e77f97a896926db9908d9cea9ed81d2758e7ba/httomo/runner/dataset_store_interfaces.py#L50-L92
- the updated code in the sweep runner is assuming the existence of the private attribute `._data` on the data source (which is questionable)
https://github.com/DiamondLightSource/httomo/blob/46e77f97a896926db9908d9cea9ed81d2758e7ba/httomo/sweep_runner/param_sweep_runner.py#L163

I would say that the best way to resolve this is to not use the private attribute, and find another way to get access to the raw data global shape. This would involve more changes and some thinking of course, but I think it's the safer way to do things - I'm not sure there's many places where reliance on private members of an object is advocated.

If accessing the private attribute absolutely must be the way to do this, then I agree that the test will need to be changed to accommodate the assumption the code is making about the existence of the private attribute `._data`. How it should be changed, I don't yet know, that will also require some thinking.

Before, the loader wrapper + loader mocks could be used because the sweep runner was not making any assumption about private members of a `DataSetSource`, so we didn't need to provide a "real" loader. This in turn made writing the test simpler, since we could use a mock loader instead of a real one. But now we need a real loader (because only a real loader has the private attribute `._data`), and the test needs to be rethought a bit.

yousefmoazzam:
As an example of a potential way to solve this without doing the private attribute access would be to modify `DataSetSource`.

_If_ it makes sense for any data source to have the information about the raw data's global shape (which should be checked if it's reasonable or not), then one _could_:
- add a `raw_shape` property to `DataSetSource` (or some other name that implies "the shape of the raw data")
- for all implementors of `DataSetSource`, add an implementation of that property
- for example, in the case of `StandardTomoLoader`, it might do something like this:
```python
@property
def raw_shape(self) -> Tuple[int, int, int]
    return self._data.shape
```
- then, the sweep runner could do `source.raw_shape` instead of `source._data.shape`, and would not be assuming the existence of any private attribute on the loader

This change would be minimal, and would avoid doing any private member access.

This is just an example to illustrate that there are ways to go about this without the private member access, and to provide some sort of substance to my advice to "find another way to get access to the raw data global shape".

yousefmoazzam (commented):
A decent start, thanks! I've made some comments on a few things, the public vs. private stuff and the lack of tests for the new feature I think are the main points (in addition to the concern about the private member access of `source._data` mentioned outside of this review).

dkazanc:
> This is just an example to illustrate that there are ways to go about this without the private member access, and to provide some sort of substance to my advice to "find another way to get access to the raw data global shape".

Thanks, I've done some changes as you suggested,  `test_execute_modifies_block`  still fails but now on `PreviewConfig` is not being available in the loader as this is something we [need](https://github.com/DiamondLightSource/httomo/blob/7cadc11b64230d700d4459054108609afbe55d50/httomo/sweep_runner/param_sweep_runner.py#L163), before it gets reassigned. Should `make_test_loader` take preview into account somehow?

httomo/sweep_runner/param_sweep_runner.py
```diff
@@ -240,3 +278,73 @@ def _get_param_sweep_name_and_vals(d: Dict[str, Any]) -> Tuple[str, Tuple[Any, .
...
+    Estimating the number of vertical slices than can fit the device to run Paganin method
+    """
+    factor = 10  # ad-hoc parameter as no real memory estimation is performed. Should be increased if OOM happens.
+    available_memory = get_available_gpu_memory(10.0)
+    available_memory_in_GB = round(available_memory / (1024**3), 2)
+    tot_slices_fit = int(
+        available_memory_in_GB
+        // (((source.aux_data.angles_length * source.global_shape[2]) * 4) / 1024**3)
+    )
+    return tot_slices_fit // factor
```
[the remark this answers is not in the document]
dkazanc: Yes, this is a one naive way to _significantly_ overestimate for the amount of slices we can fit into a GPU. It is also related not to only Paganin but rather the whole pipeline, or rather to the most memory hungry method there. For instance, if FBP (or with more certainty log-polar) will take more memory than Paganin, but we did estimation for Paganin only, we will still end up with OOM. It seems that to make this truly accurate is to go through every method in the pipeline and do memory calculations for each in order to decide on that size of the block. This seems to be quite tedious task, so I've chosen this way for now.

The other possible way  is something what we have discussed. Do GPU-specific runs of pipelines where Paganin and FBP present and some other memory hungry methods (I think for now these two are the main candidates). Estimate the **actual** number of slices before the OOM occurs and hard code it. This might introduce possibly more conditions, but I reckon it could be safer?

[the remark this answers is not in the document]
yousefmoazzam: Ok, thanks for the explanation, that makes it clearer to me what this is for.

With the paganin filter kernel sometimes requiring more slices and thus potentially causing a GPU OOM for paganin, it is indeed a bit unclear where to draw the line with memory estimation (ie, as you point out, because paganin can force lots of slices potentially, this then means other GPU methods can OOM with that many slices, so do we estimate for other methods too or not...).

Let's say we keep this approach for now, and when the feature is ready to test before merging, we see how well the approach stands up to doing sweeps on a few different GPU pipelines. If it's getting OOM in some pipelines, we may need to rehtink. If no OOM's are occurring, then we can merge, and make a new issue that details how this could be improved.

yousefmoazzam:
> Thanks, I've done some changes as you suggested, `test_execute_modifies_block` still fails but now on `PreviewConfig` is not being available in the loader as this is something we [need](https://github.com/DiamondLightSource/httomo/blob/7cadc11b64230d700d4459054108609afbe55d50/httomo/sweep_runner/param_sweep_runner.py#L163), before it gets reassigned. Should `make_test_loader` take preview into account somehow?

This is a similar issue to the previous one, but now with `StandardLoaderWrapper` (and not involving private member access, but instead involving implementation-specific assumptions):
- `self._pipeline.loader` is an object which implements `LoaderInterface` https://github.com/DiamondLightSource/httomo/blob/f46c6d66accc78b5340684d987eb6f7451a81baa/httomo/runner/pipeline.py#L16-L18
- `LoaderInterface` doesn't say anything about having a `.preview` attribute (or getter method): https://github.com/DiamondLightSource/httomo/blob/f46c6d66accc78b5340684d987eb6f7451a81baa/httomo/runner/loader.py#L7-L36
- `StandardLoaderWrapper` is an implementor of `LoaderInterface`, and it happens to have a `.preview` attribute (ie, it's implementation-specific to this loader wrapper, the `.preview` attribute is not a general part of what `LoaderInterface`): https://github.com/DiamondLightSource/httomo/blob/f46c6d66accc78b5340684d987eb6f7451a81baa/httomo/loaders/standard_tomo_loader.py#L370-L401
- so, the sweep runner is assuming things about the loader wrapper which it shouldn't assume (based on the current definition of `LoaderInterface`)

Also similar to before, a potential solution would be to modify the protocol in question (`LoaderInterface`) and its implementor (`StandardLoaderWrapper`).

_If_ it would make sense to have any "loader wrapper" provide the associated preview config (which does sound reasonable to me), then it may make sense to:
- add a `preview` getter method to `LoaderInterface`
- add an implementation of it in `StandardLoaderWrapper`, something like:
```python
@property
def preview(self) -> PreviewConfig
    return self.preview
```
- modify `make_test_loader()` to take in a `preview: PreviewConfig` parameter and pass it in to `mocker.autospec()` when defining a mock implementor of `LoaderInterface`
- in the test that's failing, update the call to `make_test_loader()` to pass in some preview config object (for the sake of consistency, making the preview config reflect what the `GLOBAL_SHAPE`/`PREVIEWED_SLICES_SHAPE` values are would make the most sense)

Side note: given that the loader wrapper would provide the preview if the above is done, and that the loader itself also would be able to provide this info, I'm becoming more wary of if the existence of the loader wrapper makes much sense (and if the wrapper should instead just go away, and loaders should be direct implementors of `LoaderInterface`). Tagging #504, as this is a relevant piece of info for that.

dkazanc:
Thanks Yousef, this gets a bit more in-depth, but I will give it a go. Also the tests will be failing even if the preview stuff is sorted, because of [this](https://github.com/DiamondLightSource/httomo/blob/fd8f6a6ed3ba96a483d4768d681faa18993ea807/httomo/sweep_runner/param_sweep_runner.py#L167) as `source.raw_shape` is not defined for `DataSetBlock`. Should it be though?

yousefmoazzam:
> Thanks Yousef, this gets a bit more in-depth, but I will give it a go. Also the tests will be failing even if the preview stuff is sorted, because of [this](https://github.com/DiamondLightSource/httomo/blob/fd8f6a6ed3ba96a483d4768d681faa18993ea807/httomo/sweep_runner/param_sweep_runner.py#L167) as `source.raw_shape` is not defined for `DataSetBlock`. Should it be though?

Sorry, I'm not sure I understand: where is the need for `DataSetBlock` to have the `.raw_shape` property? The `source` variable in that function is only ever a value of the type returned by `self._pipeline.loader.make_data_source()`, and the type that function returns is `DataSetSource`. Could you point me to where `DataSetBlock` comes into the picture here?

dkazanc:
Yes, its the tests for the sweep runs, they all build around using [DataSetBlock](https://github.com/DiamondLightSource/httomo/blob/fd8f6a6ed3ba96a483d4768d681faa18993ea807/tests/sweep_runner/test_param_sweep_runner.py#L143) rather than the loader where `source.raw_shape` would be available otherwise.

yousefmoazzam:
Ah ok, thanks for the link, I see that the tests create a `DataSetBlock` for the block splitter to return (which gets the block from the mock loader), so that's one piece of the puzzle in understanding a bit better on my end.

I'm still not sure where there's a need for `DataSetBlock` to have `.raw_shape` though? The `.raw_shape` attribute is for data sources, and a `DataSetBlock` isn't a data source, it's something that is produced by a data source.

The function/purpose of the `DataSetBlock` created in the tests is that it's a block that is produced by the block splitter (which as I mentioned before, the splitter gets from the mock loader); the mock loader will already be configured to have the `.raw_shape` attribute from the changes made earlier in the PR, so `source.raw_shape` will work in the sweep runner now. And when the splitter is asked for a block here: https://github.com/DiamondLightSource/httomo/blob/fd8f6a6ed3ba96a483d4768d681faa18993ea807/httomo/sweep_runner/param_sweep_runner.py#L188

then the `DataSetBlock` created in the test is returned by the block splitter. After that, I can't yet see where the `DataSetBlock` has any need for a `.raw_shape` attribute.

So, there's something I'm still not understanding I think: could you point me to where exactly you think the `DataSetBlock` would need `.raw_shape`?

yousefmoazzam:
> OK, so a couple of things here:
> 
> 1. I've done the `preview` introduction into `Loader` as suggested. In the new test `tests_preview_modifier_paganin` I'm passing the preview config into the `make_test_loader` and then `create_autospec`. The resulting preview in the loader is a mocked object still, but I was hoping it will be what I actually [passed](https://github.com/DiamondLightSource/httomo/blob/fc38600be7b31254f730176d41ed66293beb5331/tests/sweep_runner/test_param_sweep_runner.py#L119) to it. Could it be that `create_autospec` won't support preview classes to be passed in, as I see that other variables are just normal types?

The preview needs to be passed to the autospeccing of the `LoaderInterface`, currently it's not: https://github.com/DiamondLightSource/httomo/blob/fc38600be7b31254f730176d41ed66293beb5331/tests/testing_utils.py#L51-L64

The `self._pipeline.loader.preview` is essentially doing `StandardLoaderWrapper.preview` / `LoaderInterface.preview`, and because `create_autospec()` for `LoaderInterface` hasn't been given the preview, the mock loader wrapper doesn't know about the preview, so it won't have the preview value that was passed into `make_test_loader()`.

> 2. Secondly on `raw_shape` again. In that test above `source.raw_shape` is also a mocked object, but I need an actual shape of the raw data in order to get something meaningful from the updated [preview](https://github.com/DiamondLightSource/httomo/blob/fc38600be7b31254f730176d41ed66293beb5331/httomo/sweep_runner/param_sweep_runner.py#L171) in the test. May be I should mock it somehow as apparently it shouldn't be in `DataSetBlock`, as you pointed earlier?

Yeah that makes sense, by default the mock loader wrapper won't have an implementation of the `raw_shape` getter. Patching `.raw_shape` to return the global shape defined in the test would be a reasonable approach here I think.

In the past I've seen issues with trying to use `mocker.patch.object()` (which is how most things are patched in httomo tests - see `make_mock_repo()` in `testing_utils.py` for an example) for patching getter properties. Using [`PropertyMock`](https://docs.python.org/3/library/unittest.mock.html#unittest.mock.PropertyMock) seems to be the preferred way to mock getter properties. But of course, if you find a way to use `mocker.patch.object()` to patch the `.raw_shape` getter, feel free to do it that way.

dkazanc:
@yousefmoazzam sorry even with the changes you suggested I still cannot get correct `raw_shape` from the loader. Can you have a look please?

dkazanc:
ok, currently all tests pass for me except two `test_insert_image_save_after_sweep` and `test_insert_image_save_after_sweep2`, but it is expected as PR #523 should fix them. New test `tests_preview_modifier_paganin` tests the added functionality for calculating the new preview size for sweep runs based on the kernel size of the smoothing filter. Thanks for help @yousefmoazzam

tests/sweep_runner/test_param_sweep_runner.py
```diff
@@ -455,11 +583,12 @@ def sweep_method(data: np.ndarray, param_1: int):  # type: ignore
...
     NO_OF_SWEEPS = 4
     EXPECTED_BLOCK_SHAPE = (
-        PREVIEWED_SLICES_SHAPE[0],
+        GLOBAL_SHAPE[0],
         NO_OF_SWEEPS,
-        PREVIEWED_SLICES_SHAPE[2],
+        GLOBAL_SHAPE[2],
     )
     assert runner.block.data.shape == EXPECTED_BLOCK_SHAPE
+    assert runner._vertical_slices_preview == 7
```
yousefmoazzam, line 591: I'm not particularly fond of this assertion, because it's checking something private in the sweep runner.

I also don't think it makes sense to fix the issue by just making that attribute public, because it's an internal value that the sweep runner is calculating to deal with paganin, I don't think it's something that should be exposed.

Stepping back, what we care about as far as using the sweep runner object is concerned, is that it updates the preview of the loader wrapper correctly to account for the paganin filter's kernel size. The loader wrapper's preview is public, and is what we ultimately care about, not the internal `ParamSweepRunner._vertical_slices_preview` value.

I also noticed that if I print out `loader.preview` before and after the sweep runner execution in the test, doing this instead at line 582:
```python
print(f"loader.preview before modifying: {loader.preview}")
runner.execute()
print(f"loader.preview after modifying: {loader.preview}")
```

then the preview in the `detector_y` dimension is seemingly incorrect; the start and stop are invalid (`start=48`, while `stop=3`):
```
loader.preview before execution: PreviewConfig(angles=PreviewDimConfig(start=0, stop=180),
detector_y=PreviewDimConfig(start=50, stop=53), detector_x=PreviewDimConfig(start=0, stop=160))
loader.preview after execution: PreviewConfig(angles=PreviewDimConfig(start=0, stop=180),
detector_y=PreviewDimConfig(start=48, stop=3), detector_x=PreviewDimConfig(start=0, stop=160))
```

So, while the private `self._vertical_slices_preview` value is what is expected (the test as it currently is, passes), the final public preview value looks like it's incorrect, and it's not being caught in the test.

I could be doing something wrong, so please do double-check if the preview is wrong in this test when you run it locally!

If it turns out that the preview is indeed incorrect, I think it would be better to remove the assertion on the private `self._vertical_slices_preview`, and instead assert that `loader.preview` is what we expect it to be, which takes care of both:
- the desire to assert only on public members of objects in tests
- asserting the value that ultimately we care about is correct (the updated preview)

tests/sweep_runner/test_param_sweep_runner.py
```diff
@@ -115,20 +120,90 @@ def tests_prepare_raises_error_if_too_many_sino_slices(mocker: MockerFixture):
...
-    loader = make_test_loader(mocker, block=block)
-    m1 = make_test_method(mocker)
-    sweep_wrapper = make_test_method(mocker, param_1=(0,))
-    pipeline = Pipeline(loader=loader, methods=[m1, sweep_wrapper])
-    runner = ParamSweepRunner(pipeline, MPI.COMM_WORLD)
+    preview = PreviewConfig(
+        angles=PreviewDimConfig(start=0, stop=180),
+        detector_y=PreviewDimConfig(start=0, stop=100),
+        detector_x=PreviewDimConfig(start=0, stop=160),
+    )
```
yousefmoazzam, line 127: Could you reuse the `GLOBAL_SHAPE` or `PREVIEWED_SLICES_SHAPE` variables instead of the hardcoded numbers, for the sake of not having hardcoded numbers floating around where they're already defined in variables:
```suggestion
    preview = PreviewConfig(
        angles=PreviewDimConfig(start=0, stop=PREVIEWED_SLICES_SHAPE[0]),
        detector_y=PreviewDimConfig(start=0, stop=PREVIEWED_SLICES_SHAPE[1]),
        detector_x=PreviewDimConfig(start=0, stop=PREVIEWED_SLICES_SHAPE[2]),
    )
```

tests/sweep_runner/test_param_sweep_runner.py
```diff
@@ -115,20 +120,90 @@ def tests_prepare_raises_error_if_too_many_sino_slices(mocker: MockerFixture):
...
 
-    err_str = (
-        "Parameter sweep runs support input data containing <= 7 sinogram slices, "
-        "input data contains 10 slices"
+    mocker.patch(
+        "httomo.method_wrappers.generic.import_module", return_value=FakeModule
     )
-    assert err_str in str(e)
+
+    # Stuff take from `make_test_loader()`
```
yousefmoazzam, line 137: Sorry, when I put this comment in the snippet I sent previously, I added it to try and help you to understand the code snippet, I don't think this needs to be in the final test.

tests/sweep_runner/test_param_sweep_runner.py
```diff
@@ -115,20 +120,90 @@ def tests_prepare_raises_error_if_too_many_sino_slices(mocker: MockerFixture):
...
+        return ret
+
+    mocker.patch.object(
+        loader,
+        "make_data_source",
+        side_effect=mock_make_data_source,
+    )
+
+    # Create sweep method wrapper, passing the sweep values for the param to sweep over
+    NO_OF_SWEEPS = 2
```
yousefmoazzam, line 182: Please correct me if I'm wrong, but it seems like `NO_OF_SWEEPS` isn't being used anywhere in this test? If that's the case, could it be removed please.

tests/sweep_runner/test_param_sweep_runner.py
```diff
@@ -115,20 +120,90 @@ def tests_prepare_raises_error_if_too_many_sino_slices(mocker: MockerFixture):
...
+        output_mapping={},
+        **paganin_params,
+    )
+
+    # Define pipeline, stages, runner objects, and execute the methods in the sweep stage
+    pipeline = Pipeline(loader=loader, methods=[sweep_method_wrapper])
+    runner = ParamSweepRunner(pipeline, MPI.COMM_WORLD)
+    runner.modify_loader_preview()
+
+    assert runner._vertical_slices_preview == 21
```
yousefmoazzam, line 206: Same comment as my other one about asserting on private members of an object, I'd prefer this to not be the way we check that the new feature is doing what it's expected to do.

tests/sweep_runner/test_param_sweep_runner.py
```diff
@@ -346,6 +421,11 @@ def sweep_method(data: np.ndarray, param_1: int):  # type: ignore
         "httomo.method_wrappers.generic.import_module", return_value=FakeModule
     )
 
+    mocker.patch(
+        "httomo.sweep_runner.param_sweep_runner._preview_modifier",
+        return_value=[50, 50],
+    )
+
```
yousefmoazzam, line 428: Similar to my comments about asserting on private members of an object, I'm not overfly fond of patching private methods/functions, because they should be an implementation detail that the test shouldn't need to care about.

In this test, it's not doing paganin kernel estimation stuff to modify the preview, so would it be possible to instead patch the entire public function `ParamSweepRunner.modify_loader_preview` to leave the preview alone, rather than patching the private function `_preview_modifier`?

tests/sweep_runner/test_param_sweep_runner.py
```diff
@@ -422,6 +501,55 @@ def sweep_method(data: np.ndarray, param_1: int):  # type: ignore
         "httomo.method_wrappers.generic.import_module", return_value=FakeModule
     )
 
+    preview = PreviewConfig(
+        angles=PreviewDimConfig(start=0, stop=180),
+        detector_y=PreviewDimConfig(start=50, stop=53),
+        detector_x=PreviewDimConfig(start=0, stop=160),
+    )
```
yousefmoazzam, line 508: Similar comment to another one about having extra hardcoded values than needed, could you replace these hardcoded values with the variable `GLOBAL_SHAPE` where possible, something like the following:
```suggestion
    preview = PreviewConfig(
        angles=PreviewDimConfig(start=0, stop=GLOBAL_SHAPE[0]),
        detector_y=PreviewDimConfig(start=50, stop=53),
        detector_x=PreviewDimConfig(start=0, stop=GLOBAL_SHAPE[2]),
    )
```

tests/sweep_runner/test_param_sweep_runner.py
```diff
@@ -422,6 +501,55 @@ def sweep_method(data: np.ndarray, param_1: int):  # type: ignore
...
+
+    # Stuff take from `make_test_loader()`
```
yousefmoazzam, line 510: Same as the other comment I made, this comment can be removed.

httomo/sweep_runner/param_sweep_runner.py
```diff
@@ -124,22 +130,59 @@ def determine_stages(self) -> Stages:
...
+            self._vertical_slices_preview,
+        )
+
+        self._pipeline.loader.preview = PreviewConfig(
+            angles=self._pipeline.loader.preview.angles,
+            detector_y=PreviewDimConfig(
+                start=preview_new_start_stop[0], stop=preview_new_start_stop[1]
+            ),
+            detector_x=self._pipeline.loader.preview.detector_x,
+        )
```
yousefmoazzam, line 177: Currently I get a type error from pyright telling me:
```
Cannot assign to attribute "preview" for class "LoaderInterface"
  "PreviewConfig" is not assignable to "property"
```

The reason for this error is that, while a setter for preview has been added to `StandardLoaderWrapper`, a setter for preview hasn't been added to the `LoaderInterface` protocol. In this `modify_loader_preview()` method, all we know about `self._pipeline.loader` is that it's an object that implements `LoaderInterface`, and that protocol has no preview setter, so the type checker correctly flags this as a type error.

Could you fix this by adding a preview setter to `LoaderInterface`? (Pyright/mypy should stop complaining once that's done).

yousefmoazzam (commented):
Nice, thanks for the updates!

I think we're close, but there are a few things I'd like to be addressed before merging. The main one is stuff around the assertion on the private `self._vertical_slices_preview`, and to double-check if you also see an incorrectly modified preview as mentioned in the comments below.

tests/sweep_runner/test_param_sweep_runner.py
yousefmoazzam: Come to think of it, I feel like maybe not patching it at all would be an even better strategy.

We'd want to make sure that in the absence of paganin, the newly added feature _doesn't modify the preview at all_. If the preview _did_ get changed by the new logic for a sweep pipeline that _doesn't_ have paganin, we'd want to know about it, and mocking any of the logic for modifying the preview would mask that error.

So, I'd say to remove the patching completely, we want to know if the new feature is messing with the preview when paganin isn't present, and not patching the new logic would tell us that.

tests/sweep_runner/test_param_sweep_runner.py
dkazanc: thanks @yousefmoazzam, Yes, good point, I've modified the tests accordingly. On the wrong preview stuff. I think there is an issue because of the `GLOBAL_SHAPE = (180, 3, 160)`, but we would like to extend the `preview` to be larger (7 slices, as we set `self._vertical_slices_preview = 7`). I guess, there is no check if the global size of data is actually smaller vertically than the `preview`. So I'm thinking that we should add it. I guess 3 vertical slices is the minimal size, so we can reset the `preview` to be something between 3 and 7, if the global size of data is smaller than 7?

dkazanc: that becomes a bit fiddly may be with the indices... one possible way is to make `self._vertical_slices_preview = 5` and reject the previewed data that is smaller than that?  I think this is reasonable. Other methods should work on 5 slices well enough?

dkazanc: OK, I've fixed that issue above by adding an additional condition on the size of the global data [here](https://github.com/DiamondLightSource/httomo/blob/57b60c0d8ae273e431b3063aea2d706c0e36b208/httomo/sweep_runner/param_sweep_runner.py#L142). I also added a test when the original (raw) data is too small for the sweep to work. It is set to 5 slices, which is reasonable. If the data is larger than this amount then we can manipulate with the preview: shrink or crop, but 5 is the condition.

tests/sweep_runner/test_param_sweep_runner.py
dkazanc: yep, got rid of it

httomo/sweep_runner/param_sweep_runner.py
dkazanc: Sorry cannot see the error with `mypy` somehow, can you please give it a go?

tests/sweep_runner/test_param_sweep_runner.py
yousefmoazzam: Ok, I admittedly don't really understand why 7 wasn't good and 5 is ok, but it feels like this detail may not be that significant in the bigger picture - the private member assertions have been replaced with a relevant public member assertion, and the test passes with the changes, so this feels much nicer now.

httomo/sweep_runner/param_sweep_runner.py
yousefmoazzam: No worries, added in 516f7fb

yousefmoazzam (approved):
Much appreciated for sticking it out, it was a bit of a journey! Things look good to me, thanks again :slightly_smiling_face:

tests/sweep_runner/test_param_sweep_runner.py
dkazanc: no difference really, I just thought that 5 is probably enough for 3D processing, but it is so far purely heuristic tbh. We will see if that needs increasing with time. A good example will be iterative denoising in the sweep pipeline. This is where it becomes more obvious how many slices is sufficient enough to generalise it to a full 3D volume

PR #523, opened 2024-11-06
removing autoinsertion of rescale_to_int for sweep runs

### LINKED ISSUE #520: Autoinsertion of rescale for sweep runs
Auto insertion of the [rescale](https://github.com/DiamondLightSource/httomo/blob/f46c6d66accc78b5340684d987eb6f7451a81baa/httomo/transform_layer.py#L89) method for the `sweep` run will change the data type for all subsequent methods in the pipeline.
For instance, we have Paganin method for the sweep and then FBP. The data into FBP method will be passed with a data type defined in `rescale_to_int`, which makes the reconstruction incorrect. I guess in principle the reconstruction might work with 16bit rescaled data, but we haven't adapted the method to that.

dkazanc:
Fixes #520 

As it is stated in the issue #520 the auto-insertion of `rescale_to_int` in sweep runs where Paganin and reconstruction involved makes the reconstruction invalid as the input data for FBP is converted to integer type. This has been resolved by removing auto-insertion of `rescale_to_int` but enabling a conversion of floating point data in the [image saver](https://github.com/DiamondLightSource/httomolib/pull/20) as oppose to just break previously when the data type is not uint8/16/32. This change shouldn't have any effect on the normal runner when the auto-insertion of scaling is not performed. However, when the user doesn't  scale the data before the image saver the run will not fail, but the data won't be scaled uniformly.  

It is worth to note that when the users would like to save the rescaled projections themselves into images, the data will be rescaled into uint making processing after invalid. So, essentially we might want to restrict the number of `rescale_to_int` usages present in the pipeline (1 is the limit). 

## Checklist


dkazanc:
the errors in tests should go away once `httomolib` library is updated with this [PR](https://github.com/DiamondLightSource/httomolib/pull/20)

dkazanc:
iris is down apparently, the tests are passing for me locally and `httomolib` version has been bumped to v.2.2 with the relevant changes so I will merge this.

PR #534, opened 2024-12-11
adding average_radius parameter into the rotation wrapper

dkazanc:
This PR should be used together with the [vocentering](https://github.com/DiamondLightSource/httomolibgpu/pull/169) branch of httomolibgpu. 
The changes include adding a new parameter `average_radius` into the rotation wrapper in order to perform averaging of multiple sinograms to improve SNR and therefore CoR estimation accuracy. 

## Checklist


dkazanc:
@yousefmoazzam I've added two more tests, as you suggested above, so those situations are covered now :
1.  the value is 0: the underlying method function should be given a non-averaged sinogram
2.  the value is non-zero but within range: the underlying method function should be given an averaged sinogram, based on the block data and the value of average_radius
3.  the value is non-zero and out of range: an error should be raised

yousefmoazzam:
Thanks for the updates!

Don't worry about IRIS CI failing, it's the flakey cupy import error once more... I ran the tests locally and they're all passing, so that's fine.

I made a few tweaks to the tests that I think makes them a bit clearer, where the first commit's message explains the changes. I also added a test for the last missing case (number 1. in my post above) where I think we want the method wrapper to implicitly set the average_radius  to 0 if a value for that parameter isn't passed in.

Let me know if what I changed/added makes sense to you (am happy to rollback what I did if there's issues).

dkazanc:
thanks @yousefmoazzam , yes the changes make sense and the new test as well. I'll merge the PR and do the release of the backends as well.

PR #537, opened 2025-01-07
Decide the value of frames_per_chunk from saturation bandwidth

### LINKED ISSUE #423: chunk size benchmarks for intermediate files

ptim0626:
Fixes #423 

This PR provides an option to determine `frames_per_chunk` from the saturation bandwidth of the file system in DLS when it is passed as `-1`. It results in more frames per chunk so that the writing performance is increased because less chunks are needed. Several values about raw data chunk cache (rdcc) are modified to optimise the I/O performance of the dataset.

I observed around 60-70% speed-up (depending on the number of MPI ranks) when using `--frames-per-chunk -1` over `--frames-per-chunk 1`. The performance gain of setting rdcc is not as significant as the change in chunk size. Feel free to test this with different datasets, YMMV.

## Discussion
One thing worth to discuss is whether the choice of `--frames-per-chunk -1` means 'decide it automatically' is good or not. Potentially it can be something like `--frames-per-chunk 'auto'`. 

Another thing is the default value of `frames_per_chunk`, currently it is `1` and if compression is on, I changed it to `-1` (decide automatically) to maximise performance. Feel free to comment on this as this is somehow inconsistent.

I am not sure how frequent `httomo` is used outside DLS, as the saturation bandwidth is hard-coded to 512 MiB currently. This won't be applicable outside DLS, and I am not sure whether it is better to pass something like `--saturation-bw 512`, or set `frames_per_chunk` to 1 as a fall back, or just use a conservative estimate when outside DLS. 

## Saturation benchmark
This is the benchmark I did few months ago to estimate the saturation bandwidth (on i12, applicable to GPFS03), I use 512 MiB to try to accommodate more chunks although the saturation happens roughly about 64 MiB.

bw-i12

## Checklist

### DIFF
--- a/docs/source/howto/run_httomo.rst
+++ b/docs/source/howto/run_httomo.rst
@@ -181,7 +181,8 @@ The :code:`run` command
                                       running on
       --frames-per-chunk INTEGER RANGE
                                       Number of frames per-chunk in intermediate
-                                      data (0 = write as contiguous)  [x>=0]
+                                      data (0 = write as contiguous, -1 = decide
+                                      automatically)  [x>=-1]
       --help                          Show this message and exit.
 
 Arguments
@@ -438,4 +439,12 @@ TODO
 :code:`--frames-per-chunk`
 ~~~~~~~~~~~~~~~~~~~~~~~~~~
 
-TODO
+This flag sets the number of frames in a chunk for the intermediate file.
+
+- -1 (the default), it will be decided automatically;
+- 0, continguous storage will be used (no chunk storage);
+- >= 1, the number of frames in a chunk.
+
+For most cases the default -1 should be sufficient as the actual number of
+frames in a chunk is optimised by considering the saturation bandwidth of the
+filesystem.
--- a/docs/source/howto/run_httomo.rst
+++ b/docs/source/howto/run_httomo.rst
@@ -442,7 +442,7 @@ TODO
 This flag sets the number of frames in a chunk for the intermediate file.
 
 - -1 (the default), it will be decided automatically;
-- 0, continguous storage will be used (no chunk storage);
+- 0, contiguous storage will be used (no chunk storage);
 - >= 1, the number of frames in a chunk.
 
 For most cases the default -1 should be sufficient as the actual number of
--- a/httomo/globals.py
+++ b/httomo/globals.py
@@ -7,7 +7,9 @@ gpu_id: int = -1
 MAX_CPU_SLICES: int = (
     64  # A some random number which will be overwritten by --max-cpu_slices flag during runtime
 )
-FRAMES_PER_CHUNK: int = 1  # if given as 0, then write contiguous (no chunking); if given as -1, decide automatically
+FRAMES_PER_CHUNK: int = (
+    -1
+)  # if given as 0, then write contiguous (no chunking); if given as -1, decide automatically
 INTERMEDIATE_FORMAT: str = "hdf5"
 COMPRESS_INTERMEDIATE: bool = False
 SYSLOG_SERVER = "localhost"
--- a/httomo/globals.py
+++ b/httomo/globals.py
@@ -7,7 +7,7 @@ gpu_id: int = -1
 MAX_CPU_SLICES: int = (
     64  # A some random number which will be overwritten by --max-cpu_slices flag during runtime
 )
-FRAMES_PER_CHUNK: int = 1  # if given as 0, then write contiguous (no chunking)
+FRAMES_PER_CHUNK: int = 1  # if given as 0, then write contiguous (no chunking); if given as -1, decide automatically
 INTERMEDIATE_FORMAT: str = "hdf5"
 COMPRESS_INTERMEDIATE: bool = False
 SYSLOG_SERVER = "localhost"
--- a/httomo/methods.py
+++ b/httomo/methods.py
@@ -84,12 +84,13 @@ def setup_dataset(
 ) -> h5py.Dataset:
 
     if filetype == "hdf5":
+        DIMS = [0, 1, 2]
+        non_slicing_dims = list(set(DIMS) - set([slicing_dim]))
+
         if frames_per_chunk == -1:
             # decide the number of frames in a chunk by maximising the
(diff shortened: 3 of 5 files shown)


httomo/globals.py
```diff
@@ -7,7 +7,7 @@
 MAX_CPU_SLICES: int = (
     64  # A some random number which will be overwritten by --max-cpu_slices flag during runtime
 )
-FRAMES_PER_CHUNK: int = 1  # if given as 0, then write contiguous (no chunking)
+FRAMES_PER_CHUNK: int = 1  # if given as 0, then write contiguous (no chunking); if given as -1, decide automatically
```
dkazanc, line 10: Do I understand it correctly, there is 3 cases here:
1. 0 write contiguous (no chunking)
2. > 0 set to a desired number of chunks per frame explicitly
3. -1 automatic selection of chunks

dkazanc:
Thanks @ptim0626 ! I think it is reasonable to keep the number of chunks per frame to be `-1` as default?  

> I am not sure how frequent httomo is used outside DLS, as the saturation bandwidth is hard-coded to 512 MiB currently. This won't be applicable outside DLS, and I am not sure whether it is better to pass something like --saturation-bw 512, or set frames_per_chunk to 1 as a fall back, or just use a conservative estimate when outside DLS.

I think we should focus on performance at DLS so far, we do not yet expect a frequent usage of HTTomo outside Diamond. Just to understand, what does it mean in practice if someone runs httomo with saturation bandwidth card-coded to 512MB outside of Diamond?  I guess their bandwidth might be slower which would result in what?

httomo/globals.py
ptim0626: Yes there are 3 cases like what you described. I am just not sure overloading `frames-per-chunk` in this way is how httomo does thing.

ptim0626:
If the hard-coded value exceeds the actual saturation bandwidth by a lot, it will probably make the writing time longer. I had tried to increase this to something like 3 GiB in DLS and observed some slowing down, although not significantly.

Making the default to `-1` makes sense as I guess most of the users don't really care the chunk size, it is just this will change the default behaviour of httomo and I am not sure how important is compatibility here.

tests/test_methods.py
```diff
@@ -268,8 +268,13 @@ def test_save_intermediate_data_frames_per_chunk(
     with h5py.File(tmp_path / FILE_NAME, "r") as f:
         chunk_shape = f[DATA_PATH].chunks
 
-    if frames_per_chunk != 0:
+    if frames_per_chunk > 0:
         assert chunk_shape == tuple(expected_chunk_shape)
+    elif frames_per_chunk == -1:
+        # when this is -1, it is decided autmatically and because of the
+        # very small frame size, this exceeds the value of the slicing
+        # dimension, and the fallback is 1
```
yousefmoazzam, line 276: Is this note about how small frame sizes can influence the chunk size when automatically deciding the chunk size worth putting in the function doing the chunk size calculation, or is it only applicable to the test?

I'm basically wondering if this is a situation that could crop up often enough that it's worth having the info in the function doing the calculation, rather than tucked away in a test.

yousefmoazzam (commented):
Overall looks good to me, just a few small things to mention

docs/source/howto/run_httomo.rst
```diff
@@ -438,4 +439,12 @@ TODO
 :code:`--frames-per-chunk`
 ~~~~~~~~~~~~~~~~~~~~~~~~~~
 
-TODO
+This flag sets the number of frames in a chunk for the intermediate file.
+
+- -1 (the default), it will be decided automatically;
+- 0, continguous storage will be used (no chunk storage);
```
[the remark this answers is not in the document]
ptim0626: Thanks for spotting this!

httomo/methods.py
```diff
@@ -16,6 +16,10 @@
 # save a copy of the original guess_chunk if it needs to be restored
 ORIGINAL_GUESS_CHUNK = h5py._hl.filters.guess_chunk
 
+# the bandwidth that saturates the file system (single process)
+# this was estimated after performing some benchmarks
+SATURATION_BW = 512 * 2**20
```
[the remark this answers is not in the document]
ptim0626: Would putting a link to this PR helpful? The value was graphically determined from a heuristic approach so may refer future devs to this graph?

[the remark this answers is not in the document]
yousefmoazzam: That's a good idea: add to the code comment that it's based on GPFS03 at DLS for some easy extra info in the code itself, and can put a link to the PR for devs who wish to see more info.

tests/test_methods.py
yousefmoazzam: > For this to happen the frame will need to be several order of magnitudes smaller than 512 MiB

Gotcha, thanks for the clarification. It does seem like this information is relevant only in testing cases - which means it makes sense to leave in the test.

yousefmoazzam (approved):
Cheers, looks good! IRIS/cupy was even happy enough to run the tests properly, so we have a nice green tick for once :laughing:

PR #542, opened 2025-01-29
New testing structure with auto-pipelines generation

dkazanc:
Fixes #ISSUE

## Checklist


dkazanc:
The tests on Jenkins now in green and the framework tests pass for me locally. 

image

tests/conftest.py
```diff
@@ -174,98 +216,172 @@ def standard_image_key_path():
...
+@pytest.fixture
+def gpu_pipeline_360_paganin_FBP():
+    return "docs/source/pipelines_full/gpu_360_paganin_FBP.yaml"
+
+
+# ---------------------END------------------------#
+
+###########Raw projection data (big)##################
+# Note that raw_data folder should exist with the datasets
+# listed bellow
```
yousefmoazzam, line 278: I think this code comment could do with a bit more description about the existence of the `tests/test_data/raw_data` directory.

I believe this is referring to the fact that the following fixtures are only used in Jenkins tests, and that when those tests are executed on the Jenkins machine there's some mechanism which will download extra datasets and put them in the `raw_data/` subdirectory?

If someone were to just look at this code comment, I'd imagine they would look in the `tests/test_data/` directory in the repo and see that there's no `raw_data/` subdirectory, causing confusion.

tests/test_pipeline_big.py
```diff
@@ -0,0 +1,487 @@
...
+#         for i in range(slices):
+#             data_result[i, :, :] = h5f[path_to_data][:, index_prog, :]
+#             index_prog += step
+#         h5f.close()
+#     else:
+#         raise FileNotFoundError("File with httomolibgpu-FBP string cannot be found")
+
+# residual_im = data_gt - data_result
+# res_norm = np.linalg.norm(residual_im.flatten()).astype("float32")
+# assert res_norm < 1e-6
```
yousefmoazzam, line 487: While I think TODOs only existing in code comments isn't ideal, if that's the preferred way of keeping track of tasks then fair enough.

However, I think this one is more than just a bit questionable: it's not in a test function, it's at the module-level. If uncommented I don't know if this would get picked up correctly by pytest if it's being executed at the module-level, and is easily mistakable as scratch work. If this TODO needs to be kept in, could it be put inside a test function please?

tests/test_yaml_checker.py
```diff
@@ -147,23 +147,22 @@ def test_check_side_out_matches_ref_arg(
...
+#         "cpu1_pipeline",
+#         "gpu1_pipeline",
+#         "sweep_manual",
+#     ],
+# )
+# def test_validate_yaml_config(
+#     sample_pipelines: str, yaml_file: str, standard_data: str, expected: bool
+# ):
+#     yaml_file = sample_pipelines + yaml_file
+#     assert validate_yaml_config(yaml_file, standard_data) == expected
```
yousefmoazzam, line 168: The code comment could do with some extra information, it doesn't explain under what circumstances the test can be reintroduced, nor does it say how auto-generation of pipelines relates to this test being commented-out. I could hazard a guess that it's something to do with sweep-pipelines not being auto-generated or something, but I'd rather not have to guess if someone already knows the answer.

If the test doesn't make sense to be reintroduced after the changes that have been done in the PR, it may be better to just remove the test entirely.

tests/conftest.py
```diff
@@ -174,98 +216,172 @@ def standard_image_key_path():
...
+@pytest.fixture
+def i13_179623():
+    # 6k projections, 65gb dataset, 360 degrees scan
+    return "tests/test_data/raw_data/i13/360/179623.nxs"
+
+
+# @pytest.fixture
+# def i13_180622():
+#     # 6k projections, 65gb dataset, 360 degrees scan
+#     return "tests/test_data/raw_data/i13/360/180622.nxs"
```
yousefmoazzam, line 320: Any reason why this is commented-out? Is it old and should be removed? Does the pipeline not run for some reason (in which case, maybe linking to a github issue or something would be helpful)?

tests/conftest.py
```diff
@@ -360,3 +479,35 @@ def _load_yaml(yaml_in: str) -> PipelineConfig:
...
+                if save_result is not None:
+                    # add save_result to the list of keys
+                    opened_yaml[i]["save_result"] = save_result
+
+    with open(yaml_path, "w") as file_descriptor:
+        yaml.dump(
+            opened_yaml, file_descriptor, default_flow_style=False, sort_keys=False
+        )
+
+    return 0
```
yousefmoazzam, line 513: It looks like this function is only ever used in `test_pipeline_big.py`, and it's marked as being private in `conftest.py` - is there any reason to not put this in `test_pipeline_big.py` as a private helper function, rather than putting it in `conftest.py`?

yousefmoazzam, line 513: Any reason why this returns 0, none of the uses of function seem to use the return value?

yousefmoazzam (commented):
Thanks,  some comments here and there, mainly about code comments and commented-out code.

tests/test_pipeline_big.py
dkazanc: sorry it got confusing due to indentation. It is a part of the test above, which I decided to comment due to data being a bit weird one and not a representative case. I'll probably delete this test and references to this data altogether as I added more sensible data sample from the same beamline.

tests/conftest.py
dkazanc: Yeah, I mentioned that above. The data is a bit too complex so that stitching methods do not work well. We probably need a test for something more representative, so I added 179623 instead.

tests/test_yaml_checker.py
dkazanc: I actually re-enabled it, but it will run with `pytest.mark.small_data` marking as it will require the generated full pipelines. It passes for me.

tests/conftest.py
dkazanc: I was thinking that it will be used in both `test_pipeline_big` and `test_pipeline_small` if we want to extend testing a bit by modifying parameters and preview as I'm doing now in `test_pipeline_big`.

dkazanc: yeah, may be I had something in mind initially :) I'll remove it

yousefmoazzam: Ok, if it's planned to be imported and used in multiple files, then that's ok, just mark it as public. At the moment, the function is implied to be private by the leading underscore in its name, and so I was confused why a private function is being imported and used elsewhere.

PR #569, opened 2025-04-11
reusing flats/darks from a different scan

dkazanc:
Enables reusing flats/darks coming from a different scan by providing an additional `image_key_path` parameter.

## Checklist


dkazanc:
@yousefmoazzam I've got few test failures (6) in this branch that are related to the output filename changes. Specifically the error is: 
```
AssertionError: assert 2 == 1
 +  where 2 = len(['/tmp/pytest-of-algol/pytest-0/test_recon_method_output_filen3/some-recon.h5', '/tmp/pytest-of-algol/pytest-0/test_recon_method_output_filen3/some-recon.h5-3056533504-35197.lock'])
 
```
Have you seen this one? I reckon it counts the lock file being temporarily there.

yousefmoazzam:
> @yousefmoazzam I've got few test failures (6) in this branch that are related to the output filename changes. Specifically the error is:
> 
> ```
> AssertionError: assert 2 == 1
>  +  where 2 = len(['/tmp/pytest-of-algol/pytest-0/test_recon_method_output_filen3/some-recon.h5', '/tmp/pytest-of-algol/pytest-0/test_recon_method_output_filen3/some-recon.h5-3056533504-35197.lock'])
>  
> ```
> 
> Have you seen this one? I reckon it counts the lock file being temporarily there.

I have indeed seen this when running tests locally, but not in recent CI jobs.

If CI were to also report the error then I'd be inclined to think that a dependency's version has been updated which breaks things, but that isn't the case. So unfortunately I'm not yet sure what to make of it.

httomo/transform_loader_params.py
```diff
@@ -299,6 +300,12 @@ def parse_darks_flats(
     if isinstance(in_file, str):
         in_file = Path(in_file)
     data_path = config["data_path"] if config is not None else data_config.data_path
+    # dealing with the situation when image_key_path is not provided for flats and darks
+    if config is not None:
+        if "image_key_path" in config:
+            image_key_path = config["image_key_path"]
+        else:
+            image_key_path = None
```
yousefmoazzam, line 308: I'm trying to understand the logic, given the updated definition of the type `DarksFlatsParam`.

If `config is not None`, then `config` is a `DarksFlatsParam`, that makes sense to me.

`DarksFlatsParam` has a field `image_key_path`. The `image_key_path` field has type `Optional[str]`, so the value of it can be `None` or `str`. But the `image_key_path` field always has to be present, it can't be missing entirely from the `DarksFlatsParam`.

Because `image_key_path` must always be present in `DarksFlatsParam`, the result of `if "image_key_path" in config` has to always be true. So I don't understand when the `else` branch could ever execute?

Is it supposed to be checking if the value of the `image_key_path` field is `None` or `str`? If so, that check would be:
```python
if config["image_key_path"] is not None: # only True if `image_key_path` isn't `None`
```

rather than:
```python
if "image_key_path" in config: # always True due to definition of `DarksFlatsParam`
```

yousefmoazzam: I didn't catch this myself, it was actually pyright that brought this to my attention, it told me:
```
Type analysis indicates code is unreachable
```

for the `else` branch.

dkazanc: thanks, that is resolved now in [ecfcc20](https://github.com/DiamondLightSource/httomo/pull/569/commits/ecfcc20a5d4ae51e1b4bb27c0f6f08fbcc47234c)

httomo/darks_flats.py
```diff
@@ -43,6 +43,10 @@ class DarksFlatsFileConfig(NamedTuple):
 
     Therefore, an image key is not needed to distinguish between projection, dark-field, or
     flat-field images. For this case, the `image_key_path` should be given as `None`.
+
+    3. Dark-field and flat-field images are stored in separate files with image keys. This can be a new
+    dataset or two different dataset. In that case, the `image_key_path` parameter should be provided
+    for both flats and darks.
```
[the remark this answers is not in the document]
dkazanc: **This** actually referring to the case itself (a third one). May be this?
_Dark-field and flat-field images are stored in separate files with own unique or identical image keys. This can be a new
dataset or two different dataset. Therefore, the `image_key_path` parameter should be provided  for both flats and darks._

yousefmoazzam (approved):

PR #599, opened 2025-07-10
Darks or/and flats images ignore

### LINKED ISSUE #234: Data processing when no darks/flats are associated with the data
It's possible that in the future we would have to deal with data that has no darks/flats at all. Currenty, httomo doesn't support this option, so this feature would need to be added. This raises the question of how this situation would be presented inside httomo.

One option would be to have the `darks` and `flats` params for the loader have type `Optional[DarksFlatsFileConfig]`, and then use `None` values for both when there are no darks/flats associated with the data.

dkazanc:
Fixes #234 #220 

So far implementation when flats or darks or both can be ignored as a whole group of images (not individually). This will help with two cases: 1. When the user wants to ignore collected flats or darks or both. 2. When the data itself doesn't have flats and darks but some processing still required (e.g., distortion correction of projections only).

Implemented using the keyword `ignore` for flats and/or darks fields.
```
- method: standard_tomo
  parameters:
    rotation_angles:
    darks: ignore
    flats: ignore

```
This does not modify any structure of the block loader itself as darks or flats are generated as 2D images full of zeros or ones. Those are provided to the data reducer and to normalisation method if needed (won't have any effect). Might not be the most elegant implementation, however, processing the data without f/d is a special case and modifications of the framework itself to resolve this, seem unnecessary. 

Tests and documentation added.

## Checklist


httomo/darks_flats.py
```diff
@@ -47,11 +47,18 @@ class DarksFlatsFileConfig(NamedTuple):
...
+    (e.g., the case of applying distortion correction only).
+
+    5. Dark-field or flat-field images are available in the dataset but they need to be ignored, for instance to proceed
+    with the other types of data correction or reconstruction avoiding normalisation to d/f.
     """
 
     file: Path
     data_path: str
     image_key_path: Optional[str]
+    ignore: bool
```
yousefmoazzam, line 61: Small suggestion if you're interested, feel free to take it or leave it.

It seems like many times the `ignore` parameter is used, it's given a value of `False`, and that has caused a decent number of tests to be changed, simply to pass `ignore=False`. This parameter could be given a default value of `False` to reduce the number of changes needed to fix all tests which create a `DarksFlatsFileConfig` object but are simply passing `ignore=False` anyway.

yousefmoazzam (commented):
Overall looks good, comments are mainly about tweaking the docs.

Also, please apply code formatting where appropriate to satisfy the linting CI.

PR #616, opened 2025-08-11
Making a keyword (null) to work for image_key_path in the loader

dkazanc:
1. Making a keyword (`null`) to work for `image_key_path` in the loader by bypassing a check for a valid path (`params[key].strip("/")`) in yaml checker. 
2. Ensuring that the dummy flats/darks are created for data without d/f and the absence of the key
3. Some corrections about the agnostic `normalize_sino` as there was an ambiguity (cupy/numpy) between the `xp` object inside the function and the data passed to that function  

## Checklist


yousefmoazzam:
Fixes #617

PR #620, opened 2025-08-22
Block to cpu function changes of base_block

### LINKED ISSUE #619: CuPy (cupy->numpy) transfer function is unecessary applied to a numpy array.
For pure CPU methods running with the `gpu_enabled` run the numpy block triggers a CuPy API for data [transfer](https://github.com/DiamondLightSource/httomo/blob/13f27573d1c3f4fb1a5853abbc8d74cb906fe83f/httomo/base_block.py#L113) as in the code bellow
```
    def to_cpu(self):
        if not gpu_enabled:
            return
        self._data = xp.asnumpy(self.data, order="C")
```
It is also possible that the unnecessary copy of the numpy array happens meantime [this](https://github.com/cupy/cupy/blob/25e552d5d679dcdc6f7cc3d81310a9b265463137/cupy/__init__.py#L812). 
Logically, when the block is the CPU block already, no CuPy functions should be called on it. 

P.S. This is discovered through enabling `rescale_to_int` to be a pure CPU method and handled by the generic wrapper.

dkazanc:
Fixes #619 

## Checklist

- [ ] New and existing unit tests pass locally with my changes

### DIFF
--- a/httomo/base_block.py
+++ b/httomo/base_block.py
@@ -112,6 +112,7 @@ class BaseBlock(BlockData, BlockTransfer):
 
     def to_cpu(self):
         if not gpu_enabled or xp.get_array_module(self.data).__name__ == "numpy":
+            self._data = np.asarray(self.data, order="C")
             return
         self._data = xp.asnumpy(self.data, order="C")
 
--- a/httomo/base_block.py
+++ b/httomo/base_block.py
@@ -111,7 +111,7 @@ class BaseBlock(BlockData, BlockTransfer):
         self._data = xp.asarray(self.data, order="C")
 
     def to_cpu(self):
-        if not gpu_enabled:
+        if not gpu_enabled or xp.get_array_module(self.data).__name__ == "numpy":
             return
         self._data = xp.asnumpy(self.data, order="C")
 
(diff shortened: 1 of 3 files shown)


yousefmoazzam:
I'm curious, is the `order="C"` that's passed to `np.empty()` making any difference to whatever behaviour was observed?

We use numpy v1.26.x from what I see in the CI, and that version's docs for [`np.empty`](https://numpy.org/doc/1.26/reference/generated/numpy.empty.html#numpy.empty) say that `order` has a default value of `"C"`, so naively I would have thought that adding in `order="C"` doesn't do anything different?

dkazanc:
Thanks for the clarification @yousefmoazzam . You're, indeed, right about `default: ‘C’`for the `np.empty` function and I see what you mean about the non C-contiguous blocks. 
As I mentioned, we're converting non C-contiguous on the function side, when it is needed, so essentially no changes needed for the framework. I guess the main  change here is to prevent CuPy API not to be applied to numpy arrays. 
The last bit is a little change for `monitor` where do not need to track time for D2H/H2D for CPU methods.

yousefmoazzam (changes requested):
Apologies, I forgot about this PR!

Ok, it sounds like adding `order="C"` to the `np.empty()` calls indeed makes no difference, so could those be removed please.

PR #622, opened 2025-09-01
Communicate min block len only on first block in section

dkazanc:
Looks like a PR needed for this after the confirmation on the improvement made by Radway. 
Fixes #ISSUE

## Checklist


yousefmoazzam:
Cool, I'll have a look if a test can be added to guard against a regression, I think it may not be too difficult to do.

PR #623, opened 2025-09-01
Add httomo version comment to pipeline file copied to output dir

yousefmoazzam:
## Checklist


yousefmoazzam:
The version comment in the copied pipeline YAML file has the following form:
```yaml
# HTTomo version <VERSION>
```

As a reminder:
- for tagged commits (ie, releases) the version will be of the form `x.y.z`
- for other commits the version will be of the form `x.y.z.postN`, where `N` is the number of commits since the last tagged commit.

Feel free to suggest alternative wording for the comment :slightly_smiling_face:

PR #639, opened 2025-10-10
Adding archived templates into Documentation

dkazanc:
Adding links to the archived HTTomo version-specific  YAML files and full pipelines

Examples bellow:
image

For YAML templates: 
image

## Checklist


yousefmoazzam (commented):
If we're going to provide YAML templates and full pipeline templates for all release versions of httomo, it may make sense to automate the creation of the `.zip` files through CI and not have them under version control. But for now, this feels fine to do.

Otherwise, just a few small suggestions to fix typos in the docs.

PR #649, opened 2025-11-12
Add loader functionality to select subset of data along angular dimension

yousefmoazzam:
Minimal changes to enable the selection of a subset of the input data along the angular dimension, mainly as a tentative attempt to provide functionality for processing "continuous scan" data (in which there is a 3D dataset which contains multiple 180 degree tomography scans).

The implementation strategy here has been to transform the `PreviewConfig` object such that it reflects the selected subset _before_ it gets given to the loader, meaning that the loader is completely unaware of it being given only a subset of the hdf5 dataset that it's got a file handle to.

The loader parameter during development has been given the somewhat arbitrary name `continuous_scan_subset`, and it has two fields, `start` and `stop`. The parameter name isn't set in stone, it can be changed to anything we want, I just gave it that name to have something to work with.

## Checklist

- [x] The functionality enables the correct processing of subsets of real continuous scan data

### DIFF
--- a/docs/source/reference/loaders.rst
+++ b/docs/source/reference/loaders.rst
@@ -231,3 +231,27 @@ The data being loaded with the loader can be cropped/previewed prior to being
 passed along to the first method. The loader has the :code:`preview` parameter
 for configuring the cropping/previewing. Please see :ref:`previewing` for more
 details on previewing.
+
+Continuous Scan Subset Selection
+================================
+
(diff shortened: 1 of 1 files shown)


yousefmoazzam:
IRIS GPU runners are still experiencing issues, but the framework tests and small data pipeline tests (as well as the MPI tests that we don't run in CI) are passing locally, so I will go ahead and merge this.

PR #652, opened 2025-11-20
Iterative search method for memory estimator

mfep:
- Add support for iterative memory estimation
- Needed by https://github.com/DiamondLightSource/httomo-backends/pull/82

## Checklist

### DIFF
--- a/docs/source/pipelines/yaml.rst
+++ b/docs/source/pipelines/yaml.rst
@@ -58,19 +58,6 @@ Those pipelines consist of methods from HTTomolibgpu (GPU) and HTTomolib (CPU) b
     .. literalinclude:: ../pipelines_full/FISTA3d_tomobar.yaml
         :language: yaml
 
-.. _tutorials_pl_templates_cpu:
-
-Pipelines using TomoPy library
-------------------------------
-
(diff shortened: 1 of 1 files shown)


dkazanc (changes requested):
In HTTomo we also have another runner, a so-called sweep runner that performs a search/sweep over the set of given parameters. Because Paganin filter has a kernel of certain size we must take an area normally bigger or equal to the size of that kernel. 
So in this [function](https://github.com/DiamondLightSource/httomo/blob/5bd55e3656403de1f0c4b65552b4dfe6c06bb26b/httomo/sweep_runner/param_sweep_runner.py#L312) we try to estimate how large the vertical crop must be based on how many slices fit the GPU memory. This is still [calling](https://github.com/DiamondLightSource/httomo/blob/5bd55e3656403de1f0c4b65552b4dfe6c06bb26b/httomo/sweep_runner/param_sweep_runner.py#L23) an older memory estimator and the associated test actually fails as well. So we need to change that bit accordingly. We just need to know that with given parameters how many vertical slices [would fit](https://github.com/DiamondLightSource/httomo/blob/5bd55e3656403de1f0c4b65552b4dfe6c06bb26b/httomo/sweep_runner/param_sweep_runner.py#L158), exactly what the iterative estimator can do.  Can this be fixed please?

PR #661, opened 2025-12-08
Remove explicitly setting frames per chunk to 1

### LINKED ISSUE #659: Fragile intermediate data saver's frames-per-chunk calculation in multi-process runs can cause hanging
The calculation for the frames-per-chunk value is the following: https://github.com/DiamondLightSource/httomo/blob/8b3a2a06c596add2ae14e5da3377457acece6e1b/httomo/methods.py#L96-L127

These are the two possibilities after the initial calculation:
- `if frames_per_chunk > data.shape[slicing_dim]`
- `if frames_per_chunk > minimum_block_length`

The fact that both checks are run one after the other means that a process could potentially satisfy both conditions, which in turn means more opportunity for the `frames_per_chunk` value to be modified.

It has been observed that, when it comes to the data splitting for multi-process jobs, there is a broad range of edge cases in which the current logic can produce `frames_per_chunk` values that ultimately result in the file-writing hanging.

One example is when a process has some value `n > 1` for the `frames_per_chunk` for several blocks in a section, but then for the last block `frames_per_chunk` gets calculated as 1, and the changing of the `frames_per_chunk` value within one process appears to be one source of these "hanging bugs".

I have one note on how this logic could be simplified to reduce the possibility of hitting an edge case that causes hanging. The body of the first `if` statement will set `frames_per_chunk` to 1, but the second `if` statement can do the same if the `minimum_block_length` is 1. In other words, the first `if` statement is redundant, because its functionality can be taken care of by the second `if` statement. So, one suggestion would be to remove the first `if` statement, which indeed simplifies the logic, and reduces the number of edge cases.

yousefmoazzam:
Fixes #659
Fixes [IMGDA-674](https://jira.diamond.ac.uk/browse/IMGDA-674)

## Checklist

### DIFF
--- a/httomo/methods.py
+++ b/httomo/methods.py
@@ -107,15 +107,6 @@ def setup_dataset(
             # MPI ranks
             frames_per_chunk = SATURATION_BW // sz_per_chunk
 
-        if frames_per_chunk > data.shape[slicing_dim]:
-            warn_message = (
-                f"frames_per_chunk={frames_per_chunk} exceeds number of elements in "
-                f"slicing dim={slicing_dim} of data with shape {data.shape}. Falling "
-                "back to 1 frame per-chunk"
-            )
-            log_once(warn_message, logging.DEBUG)
-            frames_per_chunk = 1
-
         if frames_per_chunk > minimum_block_length:
             warn_message = (
                 f"frames_per_chunk={frames_per_chunk} exceeds length of smallest block "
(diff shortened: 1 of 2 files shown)


yousefmoazzam:
It was discussed in sprint planning that this was good to go.

The IRIS CI is currently failing due to a lack of a release of httomolib that contains the new `httomolib.misc.utils module`. This issue is unrelated to these changes to the frames per chunk calculation, so I'm going to go ahead and merge anyway.

PR #713, opened 2026-04-20
Injecting default parameters values into output yaml and copy distortion coefficients file

### LINKED ISSUE #311: Remove unused function `_check_yaml()` in `cli.py`
With the reintroduction of the `check` command in #201: https://github.com/DiamondLightSource/httomo/blob/ae5610a83a691436e0e7deff3a71c2af3f89568e/httomo/cli.py#L34-L46

there's no need for the function `_check_yaml()` now: https://github.com/DiamondLightSource/httomo/blob/ae5610a83a691436e0e7deff3a71c2af3f89568e/httomo/cli.py#L182-L184

dkazanc:
Fixes [issue](https://jira.diamond.ac.uk/browse/IMGDA-832) and [issue](https://jira.diamond.ac.uk/browse/IMGDA-833)

Issue: When default parameters are omitted in the original YAML pipeline, they are not propagated to the resulting output YAML file.

Resolution (with some caveats):

- To include a comment indicating the HTTomo version, the processed YAML file is reopened, the comment is added, and the file is saved again. This could potentially be handled more cleanly using ruamel.yaml, but that would introduce an additional dependency and some extra complexity to the code.
- For the !sweep and !sweeprange aliases, insertion of default parameters is not implemented. Instead, the original file is copied as-is to avoid reintroducing significant functionality from the YAML pipeline generator in backends. Additionally, sweep runs are primarily intended for parameter exploration rather than reproducibility. The absence of default params insertion is probably not that critical. 

I've added also distortion correction bit here as it is a small change but it affects the same part in cli.py. Hope its OK. 

## Checklist

### DIFF
--- a/httomo/cli.py
+++ b/httomo/cli.py
@@ -8,6 +8,7 @@ from typing import List, Optional, TextIO, Union, Any
 import yaml
 
 import click
+import shutil
 from mpi4py import MPI
 from loguru import logger
 
@@ -380,12 +381,16 @@ def initialise_output_directory(
 
     # If pipeline is a file path, copy it to output directory
     if isinstance(pipeline, Path):
+        pipeline_conf = yaml_loader(pipeline)
+        distortion_coeff_path = _get_distortion_coeff_path(pipeline_conf)
+        if distortion_coeff_path is not None:
+            shutil.copyfile(distortion_coeff_path, Path(httomo.globals.run_out_dir) / "dist_coeff.txt")
         path_to_pipeline = pipeline
         path_to_saved_pipeline = Path(httomo.globals.run_out_dir) / pipeline.name
         # if does_contain_sweep do not inject default parameters due to issue around "sweep" aliases in yaml
         if not does_contain_sweep:
             path_to_pipeline = path_to_saved_pipeline
-            pipeline_updated = _substitute_ommitted_default_values(pipeline)
+            pipeline_updated = _substitute_ommitted_default_values(pipeline_conf)
             with open(path_to_saved_pipeline, "w") as file_descriptor:
                 yaml.dump(
                     pipeline_updated,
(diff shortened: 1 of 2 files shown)


httomo/cli.py
```diff
@@ -369,7 +370,9 @@ def set_global_constants(
     httomo.globals.MAX_CPU_SLICES = max_cpu_slices
 
 
-def initialise_output_directory(pipeline: Union[Path, str]) -> None:
+def initialise_output_directory(
+    pipeline: Union[Path, str], does_contain_sweep: Optional[bool] = False
```
yousefmoazzam: The type `Optional[bool]` implies it can sometimes be `None`, but naively the code that calls this function looks like the value passed for the `does_contain_sweep` parameter is always a `bool`. Is there a reason this is `Optional[bool]` as opposed to `bool`?

yousefmoazzam: On a related note, does the `does_contain_sweep` parameter benefit from having a default value? The one place the function is called, a value for the `does_contain_sweep` parameter is passed, so it looks like the default value isn't doing much?

httomo/cli.py
```diff
@@ -389,6 +412,30 @@ def initialise_output_directory(pipeline: Union[Path, str]) -> None:
             f.write(pipeline)
 
 
+def _substitute_ommitted_default_values(
```
yousefmoazzam: Typo: "ommitted" should be "omitted"

httomo/cli.py
```diff
@@ -389,6 +412,30 @@ def initialise_output_directory(pipeline: Union[Path, str]) -> None:
...
+) -> PipelineConfig:
+    templates_conf = _get_template_yaml_conf(pipeline_conf)
+    for i, (method, template) in enumerate(zip(pipeline_conf, templates_conf)):
+        template_param_dict = template["parameters"]
+        method_params = set(method.get("parameters", {}).keys())
+        template_params = set(template_param_dict.keys())
+        omitted_params = template_params - method_params
+
+        for param in omitted_params:
+            # insert ommited parameter into the pipeline
```
yousefmoazzam, line 426: Typo: "ommitted" should be "omitted"

yousefmoazzam (commented):
Looks reasonable, thanks!

One note: the feature of putting the default values used into the copied YAML pipeline probably is something we want to include in the "user release notes", so adding the associated label to the PR would make sense.

Connected to this is the fact that this PR also includes the distortion coefficient file copy change, which I also think would make sense as a "user release note". However, the PR title is what appears in the user release notes, and the current PR title doesn't mention the distortion coefficients file copy change.

If the distortion coefficients file copy change is going to be in the same PR as the default values injection change, then I think this PR name should be changed to reflect that, so then the distortion coefficients file copy change isn't missed from the user release notes.

PR #719, opened 2026-05-19
adding a barrier after output folder creation

dkazanc:
Fixes the issue of not having the output folder created for other ranks than 0.
I went the `global_comm.Barrier()` way as at this point in the run there shouldn't be much of racing going on and we're not loosing time on synchronisation? It fixes the problem in Jenkins tests and pipelines run. 

## Checklist


yousefmoazzam (approved):
As mentioned in the team meeting, I think that adding a barrier is the correct way to resolve the race condition of the non-zero rank processes seeing the YAML file at all, or seeing the YAML file in the correct state.

Just needs the linting fixed and this is good to go, thanks! 🙂

PR #721, opened 2026-05-20
Broadcast autogenerated timestamp output dir name

yousefmoazzam:
This is to guard against the possibility that multiple processes have a different timestamp output dir name, due to one or more of the processes being slower than rank 0 and reaching the timestamp-generation line later than rank 0.

## Checklist


dkazanc (approved):
Thanks @yousefmoazzam , a nice and useful change. It certainly makes sense to broadcast the timestamp.

PR #736, opened 2026-07-08
Enable sweep runs over paganin filter to use hdf5 file as backing

yousefmoazzam:
Fixes [IMGDA-931](https://jira.diamond.ac.uk/browse/IMGDA-931)

## Checklist


httomo/sweep_runner/param_sweep_runner.py
```diff
@@ -29,6 +31,10 @@
 from httomo.preview import PreviewConfig, PreviewDimConfig
 from httomo.runner.dataset_store_interfaces import DataSetSource
 
+PAGANIN_SWEEP_VALUE_THRESHOLD = 50
```
dkazanc, line 34: ah, nice touch on the threshold. I think it is reasonable to assume for now. Tested, the switch works well.

dkazanc (approved):

PR #740, opened 2026-07-16
Injecting distortion coefficients into the method

dkazanc:
Fixes [issue](https://jira.diamond.ac.uk/browse/IMGDA-834), needs merging with this [PR](https://github.com/DiamondLightSource/httomolibgpu/pull/261) 

image

## Checklist

- [x] I have added the `user-release-note` label in order to include this PR in the "Notable

### DIFF
--- a/httomo/cli.py
+++ b/httomo/cli.py
@@ -430,7 +430,7 @@ def initialise_output_dir_copy_files_inject_defaults(
 ) -> None:
     """This function does the following:
     - creates output folder
-    - copies the distortion coefficient file (if provided)
+    - if distortion coefficient file is provided copies all parameters to yaml pipeline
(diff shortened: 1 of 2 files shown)


yousefmoazzam (commented):
One question regarding what users would want, I'm unsure.

In the case of providing a file with the distortion coefficients, I see that the file is no longer copied into the output dir. It looks like the python dict containing the params is modified which makes sense, but the YAML copied to the output dir won't have these injected coefficients values I think. If so, it looks like there's no record in the output dir of the coefficient values used for the run if a file with the coefficient values is provided, is that correct? And if so, is that OK for users?

yousefmoazzam (approved):

